PR #4, opened 2021-09-05
Update README.md

fishkerez:
make build was not working for me initially. I had to add export GO111MODULE=on before. Probably should add it to README

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ make ui
 
 ## Building Backend
 ```
-make backend
+export GO111MODULE=on && make backend
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -31,7 +31,7 @@ make ui
 
 ## Building Backend
 ```
-make backend
+export GO111MODULE=on && make backend
```
FrimIdan, line 34: https://maelvls.dev/go111module-everywhere/#go111module-with-go-116
`As of Go 1.16, the default behavior is GO111MODULE=on` - not sure that we need to add that

PR #17, opened 2021-09-13
Update issue templates

akpsgit:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,26 @@
...
+
+#### What happened:
+
+#### What you expected to happen:
+
+#### How to reproduce it (as minimally and precisely as possible):
+
+#### Anything else we need to know?:
+
+#### Environment:
```
FrimIdan, line 22: what about APIClarity version(we might need to integrate the version and commit-sha in the build phase to be able to extract it easily)? APIClarity logs?

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,26 @@
...
+- Kubernetes version (use `kubectl version`):
```
FrimIdan, line 23: `kubectl version --short`

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,26 @@
+---
```
FrimIdan, line 1: I see that Istio lately moved PR to a form (https://github.com/istio/istio/commit/7f810ef1b1c84e6e0d5442d60aaaf5d4c14e6d26)
maybe we should also do it? 
you can do validations for required fields and it looks better I think

.github/ISSUE_TEMPLATE/bug_report.md
akpsgit: Didn't put the version, as we don't have versions yet. Sure, adding the logs

FrimIdan (approved):

PR #18, opened 2021-09-13
fix: point wasm-filters submodule to https URL

kaechele:
This allows anonymous pulls of the submodule.


FrimIdan (approved):
Thank you!

PR #19, opened 2021-09-13
Add persistant volume

akpsgit:


deployment/apiclarity.yaml
```diff
@@ -103,6 +117,16 @@ spec:
             limits:
               memory: "1000Mi"
               cpu: "1000m"
+          volumeMounts:
+            - name: apiclarity
+              mountPath: /apiclarity
+              subPath:
```
FrimIdan, line 123: remove `subPath`

deployment/apiclarity.yaml
```diff
@@ -103,6 +117,16 @@ spec:
...
+      securityContext:
+        fsGroup: 65534
```
FrimIdan, line 125: shouldn't it be 1000 the same as `runAsGroup: 1000`?

FrimIdan (approved):

PR #20, opened 2021-09-14
fix(review): update speculator pointer and add print for approved path item

FrimIdan:


backend/pkg/rest/review_controller.go
```diff
@@ -122,13 +122,15 @@ func createAPIPaths(apiID uint, review *speculatorspec.ApprovedSpecReview) []*da
...
-		ret.PathItemsReview = append(ret.PathItemsReview, &speculatorspec.ApprovedSpecReviewPathItem{
+		approvedSpecReviewPathItem := &speculatorspec.ApprovedSpecReviewPathItem{
 			ReviewPathItem: speculatorspec.ReviewPathItem{
 				ParameterizedPath: reviewPathItem.SuggestedPath,
 				Paths:             createPathMap(reviewPathItem.APIEventsPaths),
 			},
 			PathUUID: uuid.NewV4().String(),
-		})
+		}
+		log.Infof("Approving path item: %+v", approvedSpecReviewPathItem)
```
akpsgit, line 132: Is %+v good enough? I think that the content should be printed (and marshalled?) in debug. WDYT?

FrimIdan: no pointers so it will be print as follow
```
time="2021-09-14T08:11:00Z" level=info msg="Approving path item: &{ReviewPathItem:{ParameterizedPath:/carts/{customerID}/merge Paths:map[/carts/61404cd74b9a2c00016e7ec8/merge:true]} PathUUID:2329795a-581b-412b-a233-63ac95d744f3}" func=github.com/apiclarity/apiclarity/backend/pkg/rest.createApprovedReviewForSpeculator file="/build/backend/pkg/rest/review_controller.go:132"
```

don't think that we need to marshal it

akpsgit: cool

akpsgit (approved):

PR #23, opened 2021-09-19
feat: add apiclarity version command and update version details during build

FrimIdan:
Output example
```
$ kubectl -n apiclarity exec deploy/apiclarity -- ./backend version
Version: latest
Commit: e5bc6fddc7308b15e46cf3c2c6144074ce168210
Build Time: 2021-09-19T06:34:49Z
```

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -25,5 +25,6 @@ assignees: ''
 #### Environment:
 - Kubernetes version (use `kubectl version --short`):
 - Istio version (use `istioctl version`):
+- APIClarity version (use `kubectl -n apiclarity exec deploy/apiclarity -- ./backend version`)
 - Cloud provider or hardware configuration:
 - Others:
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -25,6 +25,10 @@ linters-settings:
     settings:
       mnd:
         ignored-files: fake.go
+  forbidigo:
+    # Forbid the following identifiers (identifiers are written using regexp):
+    forbid:
+      - ^print.*$
 
 linters:
   enable-all: true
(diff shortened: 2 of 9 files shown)


.golangci.yml
```diff
@@ -25,6 +25,10 @@ linters-settings:
     settings:
       mnd:
         ignored-files: fake.go
+  forbidigo:
+    # Forbid the following identifiers (identifiers are written using regexp):
+    forbid:
+      - ^print.*$
```
akpsgit, line 31: what should this catch? maybe this?
^fmt\.Print.*$ -- forbid use of Print statements because they are likely just for debugging

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -25,5 +25,6 @@ assignees: ''
 #### Environment:
 - Kubernetes version (use `kubectl version --short`):
 - Istio version (use `istioctl version`):
+- APIClarity version (use `kubectl -n apiclarity exec deploy/apiclarity -- ./backend version`)
```
fishkerez, line 28: Do we want to make a CLI for apiclarity? could be nice

.golangci.yml
FrimIdan: this should catch all `printf("Test")` for example
The reason that I didn't used `^fmt.Print.*$` (it is the default) is that I use `fmt.Printf` in the version print command

.github/ISSUE_TEMPLATE/bug_report.md
FrimIdan: Also thought about it - could be nice but not sure it worth only for printing version
We can make all the installation process handled by the CLI and convert all the bash logic to code - need to think about it.

.golangci.yml
akpsgit: what is printf()? :)

FrimIdan: where do you see printf() ? :)

akpsgit (approved):

PR #21, opened 2021-09-17
Readme fixes

kaechele:
Here are some fixes for the README.
The changes should help clarify that Istio injection is required for APIClarity to work.

I also edited some of the wording and removed the redundant first paragraph, which was almost identical to the "Solution" paragraph.

This file now also successfully validates against [markdownlint-cli](https://github.com/igorshubovych/markdownlint-cli).

If you have any feedback let me know or feel free to edit this PR directly.

PR #24, opened 2021-09-19
doc: Add a comment for Istio injection

akpsgit:


kaechele:
Added a similar note to my README overhaul over in #21 to avoid creating a conflict.

PR #25, opened 2021-09-19
fix: approved review - handle error

fishkerez:


backend/pkg/rest/review_controller.go
```diff
@@ -50,7 +51,13 @@ func (s *Server) PostAPIInventoryReviewIDApprovedReview(params operations.PostAP
 
 	approvedReview := createApprovedReviewForSpeculator(params.Body, pathToPathItem)
 	// apply approved review to the speculator
-	s.speculator.ApplyApprovedReview(speculator.SpecKey(review.SpecKey), approvedReview)
+	if err := s.speculator.ApplyApprovedReview(speculator.SpecKey(review.SpecKey), approvedReview); err != nil {
+		errMsg := fmt.Sprintf("failed to apply approved review. %v", err)
```
akpsgit, line 55: Failed to apply the approved review.

backend/pkg/rest/review_controller.go
```diff
@@ -50,7 +51,13 @@ func (s *Server) PostAPIInventoryReviewIDApprovedReview(params operations.PostAP
...
+		log.Error(errMsg)
+		return operations.NewPostAPIInventoryReviewIDApprovedReviewDefault(http.StatusInternalServerError).WithPayload(&models.APIResponse{
```
FrimIdan, line 57: some errors are internal errors (ex. `failed to clone spec`)
and some are verification errors (ex. `failed to validate the spec`)
Don't you think we need to return the full message only on verification and with a different status code?

Also, `failed to validate the spec. %v spec: %s` will print the all spec - not sure it will be pretty in the UI

fishkerez: of course we should. But currently we dont have the infra for this. This will require changes in the api that were not discussed with Galia. We should discuss if we want to improve the mechanism

fishkerez: @FrimIdan please check the following commit in the speculator:
https://github.com/apiclarity/speculator/pull/8/commits/13a96f7c0ae50cae36a0c9ab64ad5b2484d07195

I removed the raw spec from the validation error message. I also added a SpecValidation error type. We can use it once we will agree with Galia on the api of how she wants to get the error type.

The returned validation error now looks like this:
`Failed to apply the approved review. failed to generate Open API Spec. failed to validate the spec. spec validation failed. validation failure list:
path /{param2} overlaps with /{params2}. spec validation failed`

FrimIdan: looks great, just need to put `SpecValidationError` into `validateRawJSONSpec` when the call to `validate.Spec` fails.

FrimIdan (approved):

akpsgit (approved):

PR #26, opened 2021-09-19
fix: all known UI routes need to be served by the same index.html

FrimIdan:
Closes https://github.com/apiclarity/apiclarity/issues/22

Since we currently only have 2 UI URLs prefixes `/events` and `/inventory` I've decided to match the UI routes and NOT to look into `asset-manifest.json` to verify it's a valid asset (as described [here](https://kimrgrey.medium.com/integration-of-create-react-app-into-golang-server-in-2020-1aff6e93ee5a))


akpsgit (approved):
OMG

PR #27, opened 2021-09-19
feat: expands provided spec refs to improve diff visibility

FrimIdan:


backend/pkg/rest/apiinventory_upload_spec_test.go
```diff
@@ -15,109 +15,51 @@
...
 	}
 	tests := []struct {
 		name    string
 		args    args
+		want    []byte
 		wantErr bool
 	}{
 		{
-			name: "json spec - valid",
+			name: "sanity",
```
akpsgit, line 46: do we want to test that no expansion, leaves it intact?

backend/pkg/rest/apiinventory_upload_spec.go
```diff
@@ -37,27 +36,32 @@ import (
...
+		log.Errorf("failed to analyze spec. Spec: %s. %v", jsonSpecBytes, err)
+		return operations.NewPutAPIInventoryAPIIDSpecsProvidedSpecBadRequest().WithPayload("Spec validation failed")
 	}
-	if err := validateJSONSpec(jsonSpec); err != nil {
-		log.Errorf("Spec validation failed. Spec: %s. %v", jsonSpec, err)
+
+	// Validates an OpenAPI 2.0 specification document.
+	err = validate.Spec(analyzed, strfmt.Default)
+	if err != nil {
+		log.Errorf("spec validation failed. %v", err)
```
fishkerez, line 60: You have a function for that (validateRawJSONSpec)
we have the same function in the speculator so if you really want to go all in..

FrimIdan: I just deleted the function because it was not needed any more here (I needed to use the `analyzed`)

akpsgit (approved):

backend/pkg/rest/apiinventory_upload_spec.go
fishkerez: oh I see....

PR #29, opened 2021-09-22
feat: replace mysql with postgresql

FrimIdan:


backend/pkg/database/apievent.go
```diff
@@ -35,20 +35,20 @@ const (
...
-	isNonAPIColumnName             = "isNonApi"
-	eventTypeColumnName            = "eventType"
+	pathIDColumnName               = "pathid"
+	statusCodeColumnName           = "statuscode"
+	sourceIPColumnName             = "sourceip"
+	destinationIPColumnName        = "destinationip"
+	destinationPortColumnName      = "destinationport"
+	hasSpecDiffColumnName          = "hasspecdiff" // hasProvidedSpecDiff || hasReconstructedSpecDiff
+	hostSpecNameColumnName         = "hostspecname"
+	newReconstructedSpecColumnName = "newreconstructedspec"
```
akpsgit, line 45: god. why not new_reconstructed_spec?
https://stackoverflow.com/questions/2878248/postgresql-naming-conventions

FrimIdan: This is a good question and I wanted to address it in the note and forgot about it.
Main reason is that I didn't wanted to do API changes or quote column name on every call (quote will allow it to stay camelCase).
We have the `sortKey` that is currently defined in the API with the column name
```
  apiEventSortKey:
    name: 'sortKey'
    description: 'Sort key'
    in: 'query'
    type: 'string'
    enum:
      - time
      - method
      - path
      - statusCode
      - sourceIP
      - destinationIP
      - destinationPort
      - hasSpecDiff
      - hostSpecName
    required: true
```

And we are using it here
```
func CreateSortOrder(sortKey string, sortDir *string) string {
	return fmt.Sprintf("%v %v", strings.ToLower(sortKey), strings.ToLower(*sortDir))
}
```

I wanted to avoid
1. Conversions from `camelCase` to `snake_case`
2. Dedicated function to convert the sort key 
3. Changing the swagger definition since it will required UI changes.

If you think it's a must I will implement option number 2 from above and will change all names to `snake_case`

akpsgit: It's not a must, but I think that it's better to at least mention this in a todo and why it works currently. I'm really in favour of the conversion function, we can also do it in a separate PR, but maybe better to do it early and not to mess the db in the upgrades later.

FrimIdan: I've use `snake_case` and implemented a conversion function.
I needed to do some swagger order changes (moved `definitions` to be first in order to use [Reusable Enums](https://swagger.io/docs/specification/2-0/enums/))

backend/pkg/database/apievent.go
```diff
@@ -62,32 +63,32 @@ type APIEvent struct {
...
-	HasProvidedSpecDiff      bool              `json:"hasProvidedSpecDiff,omitempty" gorm:"column:hasProvidedSpecDiff"`
-	HasSpecDiff              bool              `json:"hasSpecDiff,omitempty" gorm:"column:hasSpecDiff"`
-	HostSpecName             string            `json:"hostSpecName,omitempty" gorm:"column:hostSpecName" faker:"oneof: test.com, example.com, kaki.org"`
-	IsNonAPI                 bool              `json:"isNonApi,omitempty" gorm:"column:isNonApi" faker:"-"`
+	StatusCode               int64             `json:"statusCode,omitempty" gorm:"column:status_code" faker:"oneof: 200, 401, 404, 500"`
+	SourceIP                 string            `json:"sourceIP,omitempty" gorm:"column:source_ip" faker:"sourceIP"`
+	DestinationIP            string            `json:"destinationIP,omitempty" gorm:"column:destination_ip" faker:"destinationIP"`
+	DestinationPort          int64             `json:"destinationPort,omitempty" gorm:"column:destination_port" faker:"oneof: 80, 443"`
+	HasReconstructedSpecDiff bool              `json:"hasReconstructedSpecDiff,omitempty" gorm:"column:hasreconstructedspecdiff"`
```
akpsgit, line 72: looks like these two were missed

akpsgit (approved):

PR #34, opened 2021-09-27
docs: add note about PVCs and StorageClass

kaechele:
also, cosmetics: set number of "Restarts" on the `kubectl get pods` output to 0
as our software is absolutely flawless and never restarts 😉


kaechele:
Actually had multiple people (including myself) hit this snag on local test clusters of theirs. So I thought it would be worth the mention.

FrimIdan (approved):

PR #54, opened 2021-11-10
feat: Kong Plugin

fishkerez:


plugins/gateway/kong/plugin.go
```diff
@@ -0,0 +1,179 @@
...
+	apiClient := client.New(transport, strfmt.Default)
+	return &Config{
+		apiClient: apiClient,
+	}
+}
+
+func (conf Config) Response(kong *pdk.PDK) {
+	telemetry, err := createTelemetry(kong)
+	if err != nil {
+		_ = kong.Log.Err(fmt.Sprintf("Failed to create telemetry. %v", err))
```
FrimIdan, line 50: need to be consistent with the error prints - I will use **:** and not **.** before the error print with "Failed to ...**:** %v"

plugins/gateway/kong/plugin.go
```diff
@@ -0,0 +1,179 @@
...
+		return nil, fmt.Errorf("failed to get routed serivce. %v", err)
+	}
+	clientIP, err := kong.Client.GetIp()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get client ip. %v", err)
+	}
+	//clientPort, err := kong.Client.GetPort()
+	//if err != nil {
+	//	return nil, fmt.Errorf("failed to get client port. %v", err)
+	//}
```
FrimIdan, line 74: remove

plugins/gateway/kong/plugin.go
```diff
@@ -0,0 +1,179 @@
...
+	// that were present in the response from the Service (ignoring headers added
+	// by Kong itself)
+	resHeaders, err := kong.ServiceResponse.GetHeaders(-1) // default limit of 100 headers
+	if err != nil {
+		return nil, fmt.Errorf("failed to get response headers. %v", err)
+	}
+
+	telemetry := models.Telemetry{
+		DestinationAddress:   ":" + strconv.Itoa(destPort), // No destination ip for now
+		DestinationNamespace: "",
```
FrimIdan, line 122: why this is not extracted from the host?

plugins/gateway/kong/plugin.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+	//clientPort, err := kong.Client.GetPort()
+	//if err != nil {
+	//	return nil, fmt.Errorf("failed to get client port. %v", err)
+	//}
+	destPort := routedService.Port
+	host := routedService.Host
+
+	// Will get the actual path that the request was sent to, not the routed one
+	path, err := kong.Request.GetPath()
```
FrimIdan, line 79: I think we should use `GetPathWithQuery`

plugins/gateway/kong/plugin.go
```diff
@@ -0,0 +1,179 @@
...
+				TruncatedBody: false,
+				Body:          strfmt.Base64(reqBody),
+				Headers:       createHeaders(reqHeaders),
+				Version:       fmt.Sprintf("%f", version),
+			},
+			Host:   parseHost(host),
+			Method: method,
+			Path:   path,
+		},
+		RequestID: "", // TODO from where? do we need?
```
FrimIdan, line 134: not sure that we are using it - you can just generate a uuid if nothing on the request. Kong is not adding some king of req id header?

plugins/gateway/kong/plugin.go
```diff
@@ -0,0 +1,179 @@
...
+			Common: &models.Common{
+				TruncatedBody: false,
+				Body:          strfmt.Base64(resBody),
+				Headers:       createHeaders(resHeaders),
+				Version:       fmt.Sprintf("%f", version),
+			},
+			StatusCode: strconv.Itoa(statusCode),
+		},
+		Scheme:        scheme,
+		SourceAddress: clientIP + ":80",
```
FrimIdan, line 145: 80?

plugins/gateway/kong/plugin.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	return &telemetry, nil
+}
+
+func parseHost(kongHost string) string {
```
FrimIdan, line 151: please add an example for the `kongHost` (<svc-name>.<namespace>.<port>.svc)

plugins/gateway/kong/plugin.go
```diff
@@ -0,0 +1,179 @@
...
+		ret = append(ret, &models.Header{
+			Key:   header,
+			Value: values[0],
+		})
+	}
+	return ret
+}
+
+var Version = "0.2"
+var Priority = 1
```
FrimIdan: what are those?
if that is the plugin version we should set it during build with build flags like we do for the backend

plugins/gateway/kong/deploy/deploy.sh
```diff
@@ -0,0 +1,11 @@
...
+KongProxyContainerName="${KONG_PROXY_CONTAINER_NAME:-proxy}"
+KongGatewayDeploymentName="${KONG_GATEWAY_DEPLOYMENT_NAME:-kong}"
+KongGatewayIngressName="${KONG_GATEWAY_INGRESS_NAME:-demo}"
+
+kubectl apply -f kongPlugin.yaml
+
+deploymentPatch=`cat "patch-deployment.yaml" | sed "s/{{KONG_PROXY_CONTAINER_NAME}}/$KongProxyContainerName/g"`
+
+kubectl patch deployments.apps ${KongGatewayDeploymentName} --patch "$deploymentPatch"
+kubectl patch ingresses.networking.k8s.io ${KongGatewayIngressName} --patch "$(cat patch-ingress.yaml)"
```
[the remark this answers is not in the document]
fishkerez: Correct. should we get namespace for each?

plugins/gateway/kong/plugin.go
fishkerez: The priority is meaningless since we are the only plugin server
The version is also just a thing you need to pass to the server. Not sure where it is being used

plugins/gateway/kong/deploy/deploy.sh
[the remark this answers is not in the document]
FrimIdan: does the plugin CRD, deployment and Ingress object can be deployed on different namespaces and still work?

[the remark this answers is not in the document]
FrimIdan: https://stackoverflow.com/questions/59844622/ingress-configuration-for-k8s-in-different-namespaces

> Now, Ingress Controller can be deployed in any namespace and is, in fact, usually deployed in a namespace separate from your app services. It can out-of-the-box see Ingress rules in all namespaces in the cluster and will pick them up.
The Ingress rules, however, must reside in the namespace where the app that they configure reside.

So the ingress object will probably have different namespace from the controller.
Regarding the `KongPlugin` I'm not sure - you will need to try because there is also [KongClusterPlugin](https://docs.konghq.com/kubernetes-ingress-controller/2.0.x/references/custom-resources/#kongclusterplugin) so not sure if the plugin need to be deployed with the service namespace or the ingress controller namespace

plugins/gateway/kong/plugin.go
FrimIdan: > since we are the only plugin server
maybe the user has its own plugins

> Not sure where it is being used
Still I would put here the tag/commit hash from the build

plugins/gateway/kong/plugin.go
FrimIdan: I meant in all prints in the file and not only here.

plugins/gateway/kong/plugin.go
```diff
@@ -147,15 +144,19 @@ func createTelemetry(kong *pdk.PDK) (*models.Telemetry, error) {
 	return &telemetry, nil
 }
 
-func parseHost(kongHost string) string {
+// KongHost: <namespace>.<svc-name>.svc.cluster.local.80
```
FrimIdan, line 147: the format was changed? I remember seeing something else

plugins/gateway/kong/plugin.go
```diff
@@ -147,15 +144,19 @@ func createTelemetry(kong *pdk.PDK) (*models.Telemetry, error) {
...
+// convert to namespace.name
```
FrimIdan, line 148: need to convert it to `name.namespace`

plugins/gateway/kong/plugin.go
fishkerez: fixed

plugins/gateway/kong/deploy/kongPlugin.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: configuration.konghq.com/v1
+kind: KongPlugin
+metadata:
+  name: apiclarity-plugin
+plugin: kong-plugin
+config:
+  host: "apiclarity.apiclarity:9000"
```
FrimIdan, line 7: please allow it to be configured with an env var during the deploy script (like `KONG_PROXY_CONTAINER_NAME`) - it will help for the helm support

FrimIdan (approved):

PR #60, opened 2021-11-21
change default apiclarity service name

fishkerez:


plugins/gateway/tyk/deploy/patch-deployment.yaml
```diff
@@ -15,7 +15,7 @@ spec:
             fieldRef:
               fieldPath: metadata.namespace
         - name: APICLARITY_HOST
-          value: "apiclarity.apiclarity:9000"
+          value: "apiclarity-apiclarity.apiclarity:9000"
```
FrimIdan, line 18: should be configured also with an env var

FrimIdan (changes requested):

plugins/gateway/tyk/deploy/patch-deployment.yaml
fishkerez: Do you mean in values file? This PR is to fix the current behavior of the plugin deployments. Once we support plugin deployments with helm we can  add env var

FrimIdan (approved):

plugins/gateway/tyk/deploy/patch-deployment.yaml
FrimIdan: env var from script, like `UPSTREAM_TELEMETRY_HOST_NAME`

PR #72, opened 2021-12-14
feat: Gopacket telemetry source

fishkerez:


plugins/gopacket/go.mod
```diff
@@ -0,0 +1,98 @@
...
+	sigs.k8s.io/structured-merge-diff/v4 v4.1.2 // indirect
+	sigs.k8s.io/yaml v1.2.0 // indirect
+)
+
+// nolint:gomoddirectives
+replace github.com/up9inc/mizu/tap/api v0.0.0 => github.com/up9inc/mizu/tap/api v0.0.0-20211107103142-4badaadcc134
+
+// nolint:gomoddirectives
+replace github.com/up9inc/mizu/shared v0.0.0 => github.com/up9inc/mizu/shared v0.0.0-20211107103142-4badaadcc134
+
```
FrimIdan, line 93: again, why not in the `require`?

plugins/gopacket/main.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/spf13/viper"
+	"github.com/urfave/cli"
+)
+
+func main() {
+	viper.AutomaticEnv()
+
+	app := cli.NewApp()
+	app.Name = "APIClarity packet source extension"
+	app.Version = "0.1.0"
```
FrimIdan, line 31: we need to set it from the build version

plugins/gopacket/run.go
```diff
@@ -0,0 +1,247 @@
...
+	opts := &tap.TapOpts{
+		HostMode: true,
+	}
+
+	agent := &Agent{
+		apiClient: plugins.NewAPIClient(runConfig.UpstreamAddress),
+	}
+
+	// set mizu logger
+	logger.InitLoggerStderrOnly(logging.INFO)
```
FrimIdan, line 73: `logging.INFO` need to be configurable

plugins/gopacket/run.go
```diff
@@ -0,0 +1,247 @@
...
+				log.Info("Telemetry has been sent")
+			}()
+		}
+	}
+}
+
+// for each connection we will get the pod to pod communication (maybe twice if they are on different nodes) and the pod to service communication.
+// the filtering logic is that we only send to API Clarity telemetries where the client is in a monitored namespace
+// and the destination is not a pod IP (service IP or external IP). Same behaviour as the wasm filter.
+func (a *Agent) shouldFilterTelemetry(item *api.OutputChannelItem) bool {
```
FrimIdan, line 133: maybe `shouldIgnoreTelemetry` ?

plugins/gopacket/run.go
```diff
@@ -0,0 +1,247 @@
...
+	}
+	resBody, truncatedBodyRes, err := plugins.ReadBody(response.Body)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read response body: %v", err)
+	}
+
+	pathAndQuery := plugins.GetPathWithQuery(request.URL)
+
+	destinationNamespace := a.podMonitor.GetPodNamespaceByIP(item.ConnectionInfo.ServerIP)
+
```
FrimIdan, line 177: `a.podMonitor.GetPodNamespaceByIP(item.ConnectionInfo.ServerIP)` called both in `createTelemetry` and `shouldFilterTelemetry` it will be better to call it once and pass as a param into the functions

plugins/gopacket/run.go
```diff
@@ -0,0 +1,247 @@
...
+	requestID :=
+
+	return &models.Telemetry{
+		DestinationAddress:   item.ConnectionInfo.ServerIP + ":" + item.ConnectionInfo.ServerPort,
+		DestinationNamespace: destinationNamespace, // TODO if this is empty, it is a request to external. but in apiclarity we assume that empty namespace means internal
```
FrimIdan, line 182: regarding the `TODO`
```
	// Set API Info type
	if b.monitor.IsInternalCIDR(destInfo.IP) {
		apiInfo.Type = models.APITypeINTERNAL
	} else {
		apiInfo.Type = models.APITypeEXTERNAL
	}
```

So if we have a destination IP we need to classify it OK, no?

plugins/gopacket/run.go
```diff
@@ -0,0 +1,247 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to read dir: %v", err)
+	}
+	extensions = make([]*api.Extension, len(files))
+	for i, file := range files {
+		filename := file.Name()
+		extension := &api.Extension{
+			Path: path.Join("/app/extensions", filename),
+		}
+		plug, err := plugin.Open("/app/extensions/http.so")
```
FrimIdan, line 221: `"/app/extensions/http.so"` --> `path.Join("/app/extensions", filename)` , no?

plugins/gopacket/run.go
```diff
@@ -0,0 +1,247 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to open plugin: %v", err)
+		}
+		extension.Plug = plug
+		symDissector, err := plug.Lookup("Dissector")
```
FrimIdan, line 226: error is not checked

plugins/utils.go
```diff
@@ -0,0 +1,93 @@
...
+
+	httptransport "github.com/go-openapi/runtime/client"
+	"github.com/go-openapi/strfmt"
+
+	"github.com/apiclarity/apiclarity/plugins/api/client/client"
+	"github.com/apiclarity/apiclarity/plugins/api/client/models"
+)
+
+const (
+	MaxBodySize = 1000 * 1000
```
FrimIdan: gofmt

FrimIdan (changes requested):

plugins/gopacket/extensions/http/go.mod
```diff
@@ -0,0 +1,16 @@
...
+	github.com/up9inc/mizu/tap/api v0.0.0
+	golang.org/x/net v0.0.0-20211101193420-4a448f8816b3
+)
+
+require (
+	github.com/google/martian v2.1.0+incompatible // indirect
+	golang.org/x/text v0.3.7 // indirect
+)
+
+replace github.com/up9inc/mizu/tap/api v0.0.0 => github.com/up9inc/mizu/tap/api v0.0.0-20211107103142-4badaadcc134
```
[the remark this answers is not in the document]
fishkerez: So if run go mod tidy or something like this, it will be fixed on that version

plugins/Makefile
```diff
@@ -33,3 +37,8 @@ push-docker-kong: docker-kong
 push-docker-tyk: docker-tyk
 	@echo "Publishing Tyk plugin Docker image ..."
 	docker push ${DOCKER_REGISTRY}/tyk-plugin-${TYK_VERSION}:${DOCKER_TAG}
+
+.PHONY: push-docker-gopacket
+push-docker-gopacket: docker-gopacket
+	@echo "Publishing gopacket extension Docker image ..."
```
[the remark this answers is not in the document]
fishkerez: It is not a plugin in the sense that we dont build it as .so file with buildmode=plugin

plugins/gopacket/deploy/gopacket-ds.yml
```diff
@@ -0,0 +1,77 @@
...
+      labels:
+        app: apiclarity-tapper
+    spec:
+      containers:
+      - command:
+        - ./agent
+        - -i
+        - any
+        env:
+        - name: GOGC # For some reason they set the garbage collector to be much less frequent then default (100)
```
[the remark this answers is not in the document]
fishkerez: mizu

plugins/gopacket/config/config.go
```diff
@@ -0,0 +1,48 @@
...
+
+	nsToTap := viper.GetString(NamespacesToTapEnv)
+
+	if err := json.Unmarshal([]byte(nsToTap), &nsToTapSlice); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal NAMESPACES_TO_TAP: %v", nsToTap)
+	}
+
+	return &Config{
+		NamespaceToTap:  nsToTapSlice,
+		UpstreamAddress: viper.GetString(UpstreamAddressEnv),
```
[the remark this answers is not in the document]
fishkerez: the default value is being set in the deploy script already

plugins/gopacket/config/config.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+func LoadConfig() (*Config, error) {
+	var nsToTapSlice []string
+
+	nsToTap := viper.GetString(NamespacesToTapEnv)
+
+	if err := json.Unmarshal([]byte(nsToTap), &nsToTapSlice); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal NAMESPACES_TO_TAP: %v", nsToTap)
+	}
```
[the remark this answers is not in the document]
fishkerez: Good question
I tried to use GetStringSlice, the problem is that I don't know how to set a string slice in environment variable, or if it is possible. Tried to look it up but could not find a solution.

plugins/go.mod
```diff
@@ -0,0 +1,33 @@
+module github.com/apiclarity/apiclarity/plugins
+
+go 1.17
+
+require (
+	github.com/apiclarity/apiclarity/plugins/api v0.0.0-20211209155154-948acf2ec933
```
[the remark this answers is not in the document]
fishkerez: It is a bit tricky to use replace here, since you count on this that you will have the same folder structure in the Dockerfile build. In tyk, we have to rely on their folder structure since they run the build from a script, and tyk is using the plugin go.mod
This is why I dont use replace

plugins/gopacket/run.go
fishkerez: `func (m *Monitor) IsInternalCIDR(ip string) bool {
	if m == nil {
		return true
	}
	if ip == "" {
		// workaround for now, assuming the telemetry is coming from gateway
		return true
	}
	if m.nodeMonitor.IsPodCIDR(ip) {
		return true
	}
	if m.serviceMonitor.IsServiceIP(ip) {
		return true
	}

	return false
}
`

fishkerez: My bad I thought we check by empty namepsace. never mind

plugins/gopacket/run.go
fishkerez: Not really this is just turning off the mizu logs. if you dont set it you get a LOT of noise in the logs (for each packet)

plugins/go.mod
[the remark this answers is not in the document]
fishkerez: We can do utils/common package

plugins/gopacket/run.go
fishkerez: It is checked in line 228:
```
if err != nil || !ok {
			return nil, fmt.Errorf("failed to load the extension: %s", extension.Path)
		}
```

plugins/Makefile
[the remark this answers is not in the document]
FrimIdan: but it is in the `plugins` folder (:

plugins/go.mod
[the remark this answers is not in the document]
FrimIdan: > It is a bit tricky to use replace here, since you count on this that you will have the same folder structure in the Dockerfile build

What is the problem to create the same structure in the docker build? we do it in all other docker files I think (also in the backend).
Not using replace here will make build issues when you are changing the plugins api swagger

plugins/gopacket/config/config.go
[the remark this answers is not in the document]
FrimIdan: did you tried a simple space separated string, for example `"ns1 ns2 ns3"`.

plugins/gopacket/config/config.go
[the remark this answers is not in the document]
FrimIdan: but still, if the deploy script is not defined well. not sure we need to count on that.
If the value is required and we are not setting a default value, we need to at least verify it

plugins/gopacket/deploy/gopacket-ds.yml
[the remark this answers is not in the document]
FrimIdan: do we want to keep that comment?
maybe just comment with a reference from where the value was taken?

plugins/gopacket/extensions/http/go.mod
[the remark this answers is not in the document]
FrimIdan: go mod tidy should not update it I think, only if you `go get` new version.
the `require` is used to fixed the version

plugins/gopacket/run.go
FrimIdan: we will probably want to be able to set it to `WARN` to avoid any info print, wouldn't we?

plugins/gopacket/run.go
FrimIdan: so just remove the `TODO` :)

plugins/gopacket/run.go
FrimIdan: not a lovely thing to do

plugins/common/go.mod
```diff
@@ -0,0 +1,34 @@
+module github.com/apiclarity/apiclarity/plugins/common
+
+go 1.17
+
+require (
+	github.com/apiclarity/apiclarity/plugins/api v0.0.0-20211219144951-90a044c1c14f
```
FrimIdan, line 6: please use replace

plugins/gateway/kong/plugin.go
```diff
@@ -17,14 +17,14 @@ package main
 
 import (
 	"fmt"
+	"github.com/apiclarity/apiclarity/plugins/common"
```
FrimIdan, line 20: sort

plugins/Makefile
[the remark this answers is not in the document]
fishkerez: well.. maybe we can change the folder name to traffic_sources :)

plugins/gopacket/extensions/http/go.mod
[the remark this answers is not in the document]
fishkerez: Ok there is another reason why I need to use replace

> go: github.com/up9inc/mizu/shared@v0.0.0-20211107103142-4badaadcc134 requires
>         github.com/up9inc/mizu/tap/api@v0.0.0: reading github.com/up9inc/mizu/tap/api/tap/api/go.mod at revision tap/api/v0.0.0: unknown revision tap/api/v0.0.0

So they (mizu) use replace in their go.mod so I need to use replace as well

plugins/gopacket/go.mod
fishkerez: See above comment

plugins/gopacket/extensions/http/go.mod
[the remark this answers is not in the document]
fishkerez: Here I can use require

plugins/gopacket/go.mod
fishkerez: Ok there is another reason why I need to use replace

    go: github.com/up9inc/mizu/shared@v0.0.0-20211107103142-4badaadcc134 requires
    github.com/up9inc/mizu/tap/api@v0.0.0: reading github.com/up9inc/mizu/tap/api/tap/api/go.mod at revision tap/api/v0.0.0: unknown revision tap/api/v0.0.0

So they (mizu) use replace in their go.mod so I need to use replace as well

FrimIdan: OK

plugins/gopacket/deploy/gopacket-ds.yml
```diff
@@ -0,0 +1,84 @@
...
+spec:
+  selector:
+    matchLabels:
+      app: apiclarity-tapper
+  template:
+    metadata:
+      labels:
+        app: apiclarity-tapper
+    spec:
+      serviceAccountName: gopacket-taper
```
FrimIdan, line 17: `gopacket-taper` or `apiclarity-tapper`?

plugins/gopacket/config/config.go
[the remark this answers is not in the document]
fishkerez: Yeah that works :)

plugins/gopacket/main.go
fishkerez: You mean to import backend go.mod

FrimIdan: no, I mean to add a version pkg and set the values during docker build time like we are doing in the backend build.

plugins/Dockerfile.gopacket
```diff
@@ -0,0 +1,37 @@
+FROM golang:1.16.6-alpine AS builder
+
+# Set necessary environment variables needed for our image.
+ENV CGO_ENABLED=1 GOOS=linux GOARCH=amd64
+
+RUN apk add libpcap-dev gcc g++ make bash
+
+# Move to agent working directory (/agent-build).
```
FrimIdan, line 8: remove comment

plugins/gopacket/monitor/pod.go
```diff
@@ -0,0 +1,166 @@
...
+		pod := p.(*v1.Pod) //nolint:forcetypeassert
+		if pm.namespacesToTap[pod.Namespace] {
+			podIPs = append(podIPs, pod.Status.PodIP)
+		}
+	}
+	return podIPs
+}
+
+func (pm *PodMonitor) GetPodNamespaceByIP(ip string) string {
+	for _, p := range pm.podStore.List() {
```
FrimIdan, line 154: need to verify that ip is not empty

plugins/gopacket/run.go
```diff
@@ -0,0 +1,242 @@
...
+		case <-ctx.Done():
+			return
+		case item := <-outputItems:
+			go func() {
+				log.Infof("Output item received: connection info: %+v", item.ConnectionInfo)
+				serverNamespace := a.podMonitor.GetPodNamespaceByIP(item.ConnectionInfo.ServerIP)
+				if a.shouldIgnoreTelemetry(item, serverNamespace) {
+					return
+				}
+				telemetry, err := a.createTelemetry(item, serverNamespace)
```
FrimIdan, line 111: as discussed, `serverNamespace` should be deleted from `createTelemetry`.
`serverNamespace := a.podMonitor.GetPodNamespaceByIP(item.ConnectionInfo.ServerIP)` need to go into `a.shouldIgnoreTelemetry`.

FrimIdan (changes requested):

.github/workflows/docker.yml
```diff
@@ -72,4 +73,14 @@ jobs:
...
+          cache-to: type=local,dest=/tmp/.buildx-cache
+
+      - name: Build gopacket extension
+        uses: docker/build-push-action@v2
+        with:
+          context: plugins
+          tags: ghcr.io/apiclarity/gopacket-agent:latest
+          file: plugins/Dockerfile.gopacket
+          push: true
+          cache-from: type=local,src=/tmp/.buildx-cache
```
FrimIdan, line 85: please add 
```
          build-args: |
            VERSION=latest
```
like in `Build APIClarity`

.github/workflows/release.yaml
```diff
@@ -126,6 +130,16 @@ jobs:
...
 
+      - name: Build and Push gopacket extension
+        uses: docker/build-push-action@v2
+        with:
+          context: plugins
+          tags: ghcr.io/apiclarity/gopacket-agent:${{ github.event.inputs.version }}
+          file: plugins/Dockerfile.gopacket
+          push: true
+          cache-from: type=local,src=/tmp/.buildx-cache
+          cache-to: type=local,dest=/tmp/.buildx-cache
```
FrimIdan, line 141: please add 
```
          build-args: |
            VERSION=${{ github.event.inputs.version }}
```
like in `Build and Push APIClarity`

plugins/Dockerfile.kong
```diff
@@ -7,8 +7,15 @@ RUN apk add --update --no-cache gcc g++
 WORKDIR /plugins
 COPY api ./api
 
+WORKDIR /plugins/common
+
+COPY common/utils.go ./
+COPY common/go.* ./
+RUN go mod download
+
```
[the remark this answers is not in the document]
fishkerez: why?

plugins/gopacket/deploy/gopacket-ds.yml
```diff
@@ -0,0 +1,84 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  annotations:
+    deprecated.daemonset.template.generation: "1"
+  name: apiclarity-taper
```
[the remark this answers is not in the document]
fishkerez: I will go with taper :)

plugins/gopacket/monitor/pod.go
```diff
@@ -0,0 +1,166 @@
...
+				pm.updateFilteredAuthorities()
+			}
+		}
+	}
+}
+
+func (pm *PodMonitor) updateFilteredAuthorities() {
+	podIPs := pm.GetPodsIps()
+	tap.SetFilterAuthorities(podIPs)
+	pm.setIsStateChange(false)
```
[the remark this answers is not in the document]
fishkerez: If I miss the state change I will catch up with it in the next interval (5 sec)

plugins/Dockerfile.kong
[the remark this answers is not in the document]
FrimIdan: this is how you speed the build time
https://petomalina.medium.com/using-go-mod-download-to-speed-up-golang-docker-builds-707591336888

plugins/gopacket/monitor/pod.go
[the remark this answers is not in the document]
FrimIdan: you are setting it to false here `pm.setIsStateChange(false)` so you will miss it if no new changes will appear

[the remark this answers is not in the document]
akpsgit: Are we sure that the store is in sync with the ResourceEventHandlerFuncs events? maybe an event was called but the srtore is not yet in sync? maybe better to get/set the value from the handlers?
Agree about the race that if in this function a handler sets it to true after line 136 this change theoretically will never/for a long time take place. Maybe it's better to listen on a change channel instead of doing it periodically?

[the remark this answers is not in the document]
FrimIdan: > Are we sure that the store is in sync with the ResourceEventHandlerFuncs events

From the docs "NewInformer returns a Store and a controller for populating the store while also providing event notifications.", so I think that we can be sure that if an event was called the store was populate.

> Maybe it's better to listen on a change channel instead of doing it periodically

I also think it would be better, with one extra thought about the first "scan". When the store is being populated the first time we will get lots of "Add" events, we should wait until the store is in synced (using `WaitForCacheSync`) and only then call `SetFilterAuthorities`, after that we can update filter authorities on any change.

plugins/gopacket/main.go
```diff
@@ -0,0 +1,58 @@
...
+	viper.AutomaticEnv()
+
+	app := cli.NewApp()
+	app.Name = "APIClarity packet source tap"
+	app.Version = version.Version
+
+	app.Usage = ""
+	app.UsageText = ""
+	app.Action = run
+	app.Flags = []cli.Flag{
```
FrimIdan, line 44: you are missing
```
		Flags: []cli.Flag{
			cli.StringFlag{
				Name:  log_utils.LogLevelFlag,
				Value: log_utils.LogLevelDefaultValue,
				Usage: log_utils.LogLevelFlagUsage,
			},
		},
```

plugins/Makefile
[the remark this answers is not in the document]
fishkerez: will do it as part of another task

plugins/gopacket/monitor/pod.go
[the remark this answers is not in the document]
fishkerez: I don't think it is a good idea to do it on every event.
In the case of someone delete all pods, you will need to update the list of pods and setAuthorities a lot of times instead of waiting for it to finish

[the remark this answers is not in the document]
fishkerez: I will add a lock in this function

akpsgit (approved):

FrimIdan (approved):

PR #75, opened 2021-12-15
fix: tyk empty response

fishkerez:
fix: https://github.com/apiclarity/apiclarity/issues/70

### DIFF
--- a/plugins/gateway/tyk/v3.2.2/plugin.go
+++ b/plugins/gateway/tyk/v3.2.2/plugin.go
@@ -16,6 +16,7 @@
 package main
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -107,10 +108,15 @@ func createTelemetry(res *http.Response, req *http.Request) (*models.Telemetry, 
 	if err != nil {
 		return nil, fmt.Errorf("failed to read request body: %v", err)
 	}
+	// Return the content to the request body (since we read it)
+	req.Body = ioutil.NopCloser(bytes.NewBuffer(reqBody))
+
 	resBody, truncatedBodyRes, err := readBody(res.Body)
 	if err != nil {
 		return nil, fmt.Errorf("failed to read response body: %v", err)
 	}
+	// Return the content to the request body (since we read it)
+	res.Body = ioutil.NopCloser(bytes.NewBuffer(resBody))
 
 	pathAndQuery := getPathWithQuery(req.URL)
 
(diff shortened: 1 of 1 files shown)


plugins/gateway/tyk/v3.2.2/plugin.go
```diff
@@ -107,10 +108,15 @@ func createTelemetry(res *http.Response, req *http.Request) (*models.Telemetry,
...
 		return nil, fmt.Errorf("failed to read request body: %v", err)
 	}
+	// Return the content to the request body (since we read it)
+	req.Body = ioutil.NopCloser(bytes.NewBuffer(reqBody))
+
 	resBody, truncatedBodyRes, err := readBody(res.Body)
 	if err != nil {
 		return nil, fmt.Errorf("failed to read response body: %v", err)
 	}
+	// Return the content to the request body (since we read it)
```
akpsgit, line 118: response body

plugins/gateway/tyk/v3.2.2/plugin.go
```diff
@@ -107,10 +108,15 @@ func createTelemetry(res *http.Response, req *http.Request) (*models.Telemetry,
 	if err != nil {
 		return nil, fmt.Errorf("failed to read request body: %v", err)
 	}
+	// Return the content to the request body (since we read it)
```
akpsgit, line 111: Return -> Restore

plugins/gateway/tyk/v3.2.2/plugin.go
akpsgit: Return -> Restore

plugins/gateway/tyk/v3.2.2/plugin.go
```diff
@@ -107,10 +108,15 @@ func createTelemetry(res *http.Response, req *http.Request) (*models.Telemetry,
...
+	req.Body = ioutil.NopCloser(bytes.NewBuffer(reqBody))
+
 	resBody, truncatedBodyRes, err := readBody(res.Body)
 	if err != nil {
 		return nil, fmt.Errorf("failed to read response body: %v", err)
 	}
+	// Return the content to the request body (since we read it)
+	res.Body = ioutil.NopCloser(bytes.NewBuffer(resBody))
```
FrimIdan, line 119: instead of duplicating the logic is it not possible to make `readBody` restore the content? maybe a pointer to the `io.ReadCloser` will need to be sent to the function.

fishkerez: 1. body is an interface so not sure I can pass it as a pointer
2. For this to work we need to pass the request/response itself, hence open two functions instead of one
3. It is not always necessary to copy the body back, since in some cases (like the gopacket plugin) we dont need the body after we read it

FrimIdan (approved):

PR #79, opened 2022-01-19
feat: Kong plugin - add request/response timestamps and support forwarded IP

fishkerez:


plugins/gateway/kong/plugin.go
```diff
@@ -65,13 +73,21 @@ func createTelemetry(kong *pdk.PDK) (*models.Telemetry, error) {
...
+	responseTime, err := common.GetTimeNowRFC3339Nano()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get time now: %v", err)
+	}
 	routedService, err := kong.Router.GetService()
 	if err != nil {
 		return nil, fmt.Errorf("failed to get routed serivce: %v", err)
 	}
-	clientIP, err := kong.Client.GetIp()
+	clientIP, err := kong.Client.GetForwardedIp()
```
FrimIdan, line 88: We don't want to use `GetIp()` if failed to get (or `ForwardedIp` is empty?)

plugins/gateway/kong/plugin.go
```diff
@@ -39,6 +40,12 @@ func New() interface{} {
 	return &Config{}
 }
 
+func (conf Config) Access(kong *pdk.PDK) {
+	if err := kong.Ctx.SetShared(common.RequestTimeContextKey, time.Now().Format(time.RFC3339Nano)); err != nil {
```
FrimIdan, line 44: can't we use `common.GetTimeNowRFC3339Nano()` and set time.Time on the shared context?

plugins/gateway/kong/README.md
```diff
@@ -28,3 +28,10 @@
        helm upgrade --values values.yaml --create-namespace apiclarity apiclarity/apiclarity -n apiclarity --install
        ```
         * A post install job will execute the installation script in your cluster
+
+### Preserving Client IP Address
+Kong is usually deployed behind a Load Balancer (using a Kubernetes Service of type LoadBalancer).
+This can result in loss of actual Client IP address and Kong observing the IP address of the Load Balancer
```
akpsgit, line 34: This can result in a loss of actual Client IP address, as Kong will get the IP address of the Load Balancer
as the Client IP address.

plugins/gateway/kong/plugin.go
fishkerez: Not sure it is needed. From what I saw, it will always return an IP even if nothing is configured and the IP is the IP of the load-balancer/node,

plugins/gateway/kong/plugin.go
fishkerez: It is redundant IMO. converting time to string and then to time and then to string again...

plugins/gateway/kong/plugin.go
```diff
@@ -65,13 +73,21 @@ func createTelemetry(kong *pdk.PDK) (*models.Telemetry, error) {
...
+	}
 	routedService, err := kong.Router.GetService()
 	if err != nil {
 		return nil, fmt.Errorf("failed to get routed serivce: %v", err)
 	}
-	clientIP, err := kong.Client.GetIp()
+	clientIP, err := kong.Client.GetForwardedIp()
 	if err != nil {
-		return nil, fmt.Errorf("failed to get client ip: %v", err)
+		return nil, fmt.Errorf("failed to get client forwarded ip: %v", err)
```
akpsgit, line 90: do we want to fail the telemetry on that error? why this error can happen?

plugins/gateway/kong/plugin.go
FrimIdan: Unless the docs claims that `GetForwardedIp()` always return an IP, I think we can just add it to be safe. Maybe someone will deploy it differently.

plugins/gateway/kong/plugin.go
FrimIdan: I just wanted to save `time.Time` as is on the `Shared` - or you can just save string here?

plugins/gateway/kong/plugin.go
fishkerez: > // kong.Client.GetForwardedIp() returns the remote address of the client making the request.
> // Unlike kong.client.get_ip, this function will consider forwarded addresses in cases
> // when a load balancer is in front of Kong. Whether this function returns a forwarded
> // address or not depends on several Kong configuration parameters:
> //
> //   - trusted_ips
> //   - real_ip_header
> //   - real_ip_recursive

The important thing being:
_

> Unlike kong.client.get_ip, this function will consider forwarded addresses in cases
> when a load balancer is in front of Kong.

_

Consider sounds to me like it does the same as GetIP + considering few other things

plugins/gateway/kong/plugin.go
fishkerez: Just save string

plugins/gateway/kong/plugin.go
fishkerez: IMO we should not fail the telemetry on that error. No idea what can cause this

plugins/gateway/kong/plugin.go
FrimIdan: sounds good

akpsgit (approved):

PR #81, opened 2022-02-01
agigee integration

invent360:
agigee integration

### DIFF
--- a/plugins/gateway/apigee/README.md
+++ b/plugins/gateway/apigee/README.md
@@ -1,0 +1,28 @@
+# Apigee policy and telemetry visualization
+ 
+## Description
+This section describes the steps used in setting up apigee proxy, creating a policy and visualizing the end results in apiclarity.
+ 
+ 
+## Starting condition
+* A running deployment of APIClarity with a service to expose it through LoadBalancer so as to publish traffic from Apigee policy, also make a note of the LB IP address.
+ The APIClarity telemetry url should be something like <API_CLARITY_SERVICE:9000/api/telemetry>. This service endpoint will be used in the javascript policy.
+ N/B: Make sure not to change the existing apicalrity service cluster type from ClusetrIP to LoadBalancer, rather make a copy of the original apiclarity and define cluster type as Load Balancer
+* A running k8s cluster where backend service to be exposed from outside will reside.
+* URL of backend service (BACKEND_SERVICE_URL) is noted.
+* An Apigee instance is currently provisioned with endpoint to gateway (APIGEE_GATEWAY_URL) noted
+ 
+## Instructions
+* Create a simple passthrough proxy which exposes the backend service (with url BACKEND_SERVICE_URL) on apigee
+* Add the BACKEND_SERVICE_URL of the backend service as target endpoint on the passthrough proxy created above.
+* Click on "Developer" Tab on Apigee UI
+* Select "PreFlow" under Target endpoints -> click the Step+ button under response area
+* Scroll down the popup window and select Javascript as policy
+* Enter policy name (or leave all as default) and click on Add
+* Past the content of the "github.com/apiclarity/plugins/gateway/apigee/telemetry-policy.js" into the newly created javascript policy
+* Replace <API_CLARITY_SERVICE:9000/api/telemetry> with noted above
+ 
+## Debugging
+* Start a debugging session by clicking on the "Debug" Tap.
+* Select the latest entry (e.g. eval-xxx) in the Env. drop down menu and click on the Start Debug Session
+* Tail the logs of the apiclarity pod, after setting debugging to true on the apiclarity deployment.yaml
--- a/plugins/gateway/apigee/telemetry-policy.js
+++ b/plugins/gateway/apigee/telemetry-policy.js
@@ -1,0 +1,117 @@
+// REQUEST
+//------>Header<------//
+var requestHeaderList = [];
+
+var requestHeaders = context.getVariable("request.headers.names"),
+
+requestResults = {};
+    
+requestHeaders = requestHeaders + '';
+
+requestHeaders = requestHeaders.slice(1, -1).split(', ');
+
+requestHeaders.forEach(function(k){
+  var v = context.getVariable("request.header." + k );
+  requestResults[k] = v;
+  var header = { key : k, value: v};
+  requestHeaderList.push(header);
+  
+});
+//------>Body<------//
+var reqContent = context.getVariable('request.content')
+
+if(reqContent !== null){
+    reqContent = crypto.base64(crypto.asBytes(reqContent))
+}
+
+
+var requestCommon = { 
+                        truncatedBody : true, 
+                        body: reqContent, 
+                        headers: requestHeaderList, 
+                        version: "0.0.1"
+                    };
+                    
+var req = { 
+            common : requestCommon, 
+            host: context.getVariable("request.header.host"), 
+            method: context.getVariable('request.verb'), 
+            path: context.getVariable("request.uri")
+          };
+
+
+// RESPONSE
+//------>Header<------//
+var responseHeaderList = [];
+
+var responseHeaders = context.getVariable("response.headers.names"),
+
+responseResults = {};
+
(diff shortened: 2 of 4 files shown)


plugins/gateway/apigee/telemetry-policy.js
```diff
@@ -0,0 +1,117 @@
...
+                        body: reqContent, 
+                        headers: requestHeaderList, 
+                        version: "0.0.1"
+                    };
+                    
+var req = { 
+            common : requestCommon, 
+            host: context.getVariable("request.header.host"), 
+            method: context.getVariable('request.verb'), 
+            path: context.getVariable("request.uri")
```
FrimIdan, line 39: Does `request.uri returns path with query? (we need to get path with query)

plugins/gateway/apigee/telemetry-policy.js
```diff
@@ -0,0 +1,117 @@
...
+
+responseResults = {};
+
+responseHeaders = responseHeaders + '';
+
+responseHeaders = responseHeaders.slice(1, -1).split(', ');
+
+responseHeaders.forEach(function(k){
+  var v = context.getVariable("response.header." + k );
+  responseResults[k] = v;
```
FrimIdan, line 57: again, `responseResults` what is it for?

plugins/gateway/apigee/telemetry-policy.js
```diff
@@ -0,0 +1,117 @@
...
+
+//------>Body<------//
+var resContent = context.getVariable('response.content')
+
+if(resContent !== null){
+    resContent = crypto.base64(crypto.asBytes(resContent))
+}
+
+var responseCommon = { 
+                        truncatedBody : true,
```
FrimIdan, line 70: again, why `true`?
need to be `true` only if body len is gte `MaxBodySize`

plugins/gateway/apigee/telemetry-policy.js
```diff
@@ -0,0 +1,117 @@
...
+  responseResults[k] = v;
+  var header = { key : k, value: v};
+  responseHeaderList.push(header);
+});
+
+//------>Body<------//
+var resContent = context.getVariable('response.content')
+
+if(resContent !== null){
+    resContent = crypto.base64(crypto.asBytes(resContent))
```
FrimIdan, line 66: body len need to be limited.
example https://github.com/apiclarity/apiclarity/blob/master/plugins/gateway/kong/plugin.go#L70

plugins/gateway/apigee/telemetry-policy.js
```diff
@@ -0,0 +1,117 @@
...
+
+
+var telemetry = {
+                    destinationAddress: ":80",
+                    destinationNamespace: "",
+                    request: req,
+                    requestID: context.getVariable("request.header.x-request-id"),
+                    response: res,
+                    scheme: context.getVariable('client.scheme'),
+                    sourceAddress: ":80"
```
FrimIdan, line 89: why `destinationAddress`, `destinationNamespace` and `sourceAddress` are empty?

invent360: because it is not running in kubernetes, my understanding is that these values are only required when it is running in a k8 cluster. This approach integrates apigee running as SAAS

plugins/gateway/apigee/telemetry-policy.js
invent360: redundant code, so removed

plugins/gateway/apigee/telemetry-policy.js
```diff
@@ -0,0 +1,117 @@
...
+
+requestResults = {};
+    
+requestHeaders = requestHeaders + '';
+
+requestHeaders = requestHeaders.slice(1, -1).split(', ');
+
+requestHeaders.forEach(function(k){
+  var v = context.getVariable("request.header." + k );
+  requestResults[k] = v;
```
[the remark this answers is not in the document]
invent360: redundant code, so removed

plugins/gateway/apigee/README.md
```diff
@@ -0,0 +1,28 @@
+# Apigee policy and telemetry visualization
+ 
+## Description
+This section describes the steps used in setting up apigee proxy, creating a policy and visualizing the end results in apiclarity.
+ 
+ 
+## Starting condition
+* A running deployment of APIClarity with a service to expose it through LoadBalancer so as to publish traffic from Apigee policy, also make a note of the LB IP address.
+ The APIClarity telemetry url should be something like <API_CLARITY_SERVICE:9000/api/telemetry>. This service endpoint will be used in the javascript policy.
+ N/B: Make sure not to change the existing apicalrity service cluster type from ClusetrIP to LoadBalancer, rather make a copy of the original apiclarity and define cluster type as Load Balancer
```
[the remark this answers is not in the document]
FrimIdan: I understand it, my point is that `apigee` is a new traffic source (external source but still a traffic source) so the user will define in the helm chart that he is going to use `apigee` - that helm value will be used to understand whether to deploy the LB service.
Meaning `plugins/gateway/apigee/k8s/apiclarity-svc.yaml` should moved to `https://github.com/apiclarity/apiclarity/tree/master/charts/apiclarity/templates` and use `{{- if .Values.trafficSource.apigee.enabled }}` to deploy it.

plugins/gateway/apigee/telemetry-policy.js
FrimIdan: the destination is deplyoed in the cluster so you can/should popolate

plugins/gateway/apigee/README.md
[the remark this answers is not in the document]
invent360: ok got it, thanks

plugins/gateway/apigee/telemetry-policy.js
invent360: yes it does

charts/apiclarity/templates/apiclarity-svc.yaml
```diff
@@ -0,0 +1,25 @@
+{{- if .Values.trafficSource.apigee.enabled }}
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ include "apiclarity.name" . }}-svc
```
FrimIdan, line 5: please change `-svc` to `-lb` and also change the file name to `lb-service`

charts/apiclarity/templates/apiclarity-svc.yaml
```diff
@@ -0,0 +1,25 @@
...
+spec:
+  ports:
+    - name: trace-server
+      port: 9000
+      protocol: TCP
+      targetPort: 9000
+    - name: backend
+      port: 8080
+      protocol: TCP
+      targetPort: 8080
```
FrimIdan, line 20: Do we need to expose also the backend port or only the trace server?

plugins/gateway/apigee/telemetry-policy.js
```diff
@@ -0,0 +1,122 @@
...
+                    };
+
+var res = {
+            common : responseCommon,
+            statusCode: context.getVariable('response.status.code').toString()
+          };
+
+
+var telemetry = {
+                    destinationAddress: context.getVariable("request.header.host"),
```
FrimIdan, line 90: APIClarity is expecting an address in the format of `ip:port` - don't you have the destination IP from the request?
If the IP is missing to `:port` will be ok

plugins/gateway/apigee/telemetry-policy.js
```diff
@@ -0,0 +1,122 @@
...
+                    destinationNamespace: "",
+                    request: req,
+                    requestID: context.getVariable("request.header.x-request-id"),
+                    response: res,
+                    scheme: context.getVariable('client.scheme'),
+                    sourceAddress: context.getVariable("request.header.x-forwarded-for")
```
FrimIdan, line 96: APIClarity is expecting an address in the format of `ip:port` - if `x-forwarded-for` is missing `:` will be ok.
You can see examples in other plugins support https://github.com/apiclarity/apiclarity/blob/master/plugins/gateway/kong/plugin.go#L172

PR #86, opened 2022-02-23
feat: Enable multi-arch builds

kaechele:
AMD64, ARM64 and ARMv7 for now.
The Tyk plugin depends on the tykio/tyk-plugin-compiler image which does
not support any other architecture than AMD64 for now.
See https://github.com/TykTechnologies/tyk-build-env/issues/4


.github/workflows/ci.yml
```diff
@@ -32,8 +39,66 @@ jobs:
...
+    - name: Get current timestamp
+      id: timestamp
+      run: echo "::set-output name=timestamp::$(date -u +'%Y-%m-%dT%H:%M:%SZ')"
+
+    - name: Build APIClarity
+      uses: docker/build-push-action@v2
+      with:
+        context: .
+        platforms: linux/amd64,linux/arm64,linux/arm/v7
+        tags: ghcr.io/apiclarity/apiclarity:${{ github.sha }}
```
FrimIdan, line 51: @kaechele I don't think we should push on every PR. We can use the 'push false' option.

Dockerfile
```diff
@@ -1,31 +1,40 @@
-FROM node:14-slim AS site-build
+## Build Frontend
+FROM --platform=$BUILDPLATFORM node:14-slim AS site-build
```
kaechele, line 2: The `--platform=$BUILDPLATFORM` tells the build process to utilize the host platform, instead of emulating the target platform. This is great for build steps like these that are not architecture dependent.

Dockerfile
```diff
@@ -1,31 +1,40 @@
...
 COPY ui .
-RUN npm i
 RUN npm run build
 
+## Build Backend
+# Cross-compilation tools
+FROM --platform=$BUILDPLATFORM tonistiigi/xx AS xx
 
-FROM golang:1.16.6-alpine AS builder
+FROM --platform=$BUILDPLATFORM golang:1.16.6-alpine AS builder
```
kaechele, line 17: Go can be easily cross-compiled, which is what we will be doing here.

kaechele:
This can wait for now. Istio doesn't support ARM64 either, I figured out :P

akpsgit:
> This can wait for now. Istio doesn't support ARM64 either, I figured out :P

bummer :( maybe we can try the Tap plugin instead of Istio in this case

the-wondersmith:
@kaechele Istio now supports multi-arch builds (they work _delightfully_ on my little RPi cluster 😅). Any chance that pushes the merge of this PR forward?

kaechele:
I'd be happy to take another look at this, yes.

eti-tme-tim:
Can we put a conditional in place to skip Tyk plugin build for non-amd64 architectures and list in the caveats that Tyk support is only available on amd64 platforms?

kaechele:
The skipping should be in place already:
For the GitHub actions:
 - https://github.com/openclarity/apiclarity/pull/86/files#diff-3f5366f6d6df3ec1179e5efadc6f350bfa88eebf4e2da589b4d94ccb85ae5e94R82
 - https://github.com/openclarity/apiclarity/pull/86/files#diff-e426ed45842837026e10e66af23d9c7077e89eacbe6958ce7cb991130ad05adaR147

And when building through the Makefile: https://github.com/openclarity/apiclarity/pull/86/files#diff-7eba0837a02c2deea05123583489cc7941f60b687718194ad0b30077c3181c62R22

Also, not sure why the build is failing in CI. Works fine locally.

PR #96, opened 2022-03-11
Adding additional values for Kong deployments

jakparks:
- Added "rbac.create" boolean option. If set to "false" will disable creation of all CR/CRBs. Defaults to true
- Added "serviceAccount.create" boolean option. If set to "false" the service account will not be created. Defaults to true
- Added "resources" value. Allows overriding default resource limits/requests for all api-clarity containers.
- Added "trafficSource.kong.patch" boolean option. If set to false the postinstall Job will be skipped entirely and only the api-clarity containers will be installed. Defaults to true.

### DIFF
--- a/charts/apiclarity/templates/deployment.yaml
+++ b/charts/apiclarity/templates/deployment.yaml
@@ -19,7 +19,7 @@ spec:
       annotations:
         sidecar.istio.io/inject: "false"
     spec:
-      serviceAccountName: {{ include "apiclarity.name" . }}
+      serviceAccountName: {{ include "apiclarity.serviceAccountName" . }}
       initContainers:
         - name: '{{ include "apiclarity.name" . }}-wait-for-db'
           image: docker.io/bitnami/postgresql:11.13.0-debian-10-r33
@@ -27,6 +27,8 @@ spec:
             do echo waiting for database; sleep 2; done;']
           securityContext:
             runAsUser: 1001
+          resources:
+          {{- toYaml .Values.apiclarity.initResources | nindent 12 }}
       containers:
         - name: apiclarity
           image: '{{ .Values.global.docker.registry }}/apiclarity:{{ .Values.apiclarity.docker.imageTag }}'
@@ -89,12 +91,7 @@ spec:
             allowPrivilegeEscalation: false
             readOnlyRootFilesystem: true
           resources:
-            requests:
-              memory: "200Mi"
-              cpu: "100m"
-            limits:
-              memory: "1000Mi"
-              cpu: "1000m"
+          {{- toYaml .Values.apiclarity.resources | nindent 12 }}
           volumeMounts:
             - name: {{ include "apiclarity.name" . }}
               mountPath: '/apiclarity'
--- a/charts/apiclarity/templates/serviceaccount.yaml
+++ b/charts/apiclarity/templates/serviceaccount.yaml
@@ -1,7 +1,9 @@
+{{- if .Values.apiclarity.serviceAccount.create -}}
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  name: {{ include "apiclarity.name" . }}
+  name: {{ include "apiclarity.serviceAccountName" . }}
   namespace: '{{ .Release.Namespace }}'
   labels:
     {{ include "apiclarity.labels" . }}
+{{- end -}}
--- a/charts/apiclarity/values.yaml
+++ b/charts/apiclarity/values.yaml
@@ -43,6 +43,35 @@ apiclarity:
   ## Logging level (debug, info, warning, error, fatal, panic).
   logLevel: warning
 
+  ## Enable/disable rbac resource creation (i.e. ClusterRole, ClusterRoleBinding)
+  rbac:
+    create: true
+
+  ## ServiceAccount settings
+  serviceAccount:
+    ## Enable/disable ServiceAccount creation. Set false to use a pre-existing account
+    create: true
+    ## Override name of ServiceAccount
+    # name:
+
+  ## Resource limits for APIClarity deployment
+  resources:
+    requests:
+      memory: "200Mi"
+      cpu: "100m"
+    limits:
+      memory: "1000Mi"
+      cpu: "1000m"
+
+  ## Resource limits for APIClarity init container deployment
+  initResources:
+    requests:
+      memory: "200Mi"
+      cpu: "100m"
+    limits:
+      memory: "1000Mi"
+      cpu: "200m"
+
 ## End of APIClarity Values
 #######################################################################################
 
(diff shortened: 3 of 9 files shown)


FrimIdan:
Hi @jakparks thank you for submitting a PR.
Can you please elaborate why `rbac.create` and `serviceAccoutn.create` is needed and how the application will work otherwise? Is it counting on the default service account?

charts/apiclarity/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: apiclarity
 appVersion: latest
 description: Charts for APIClarity deployment. Reconstruct OpenAPI Specifications from real-time workload traffic seamlessly.
-version: latest
+version: v0.9.0
```
FrimIdan, line 5: please revert

charts/apiclarity/templates/deployment.yaml
```diff
@@ -27,6 +27,8 @@ spec:
             do echo waiting for database; sleep 2; done;']
           securityContext:
             runAsUser: 1001
+          resources:
+          {{- toYaml .Values.apiclarity.resources | nindent 13 }}
```
FrimIdan, line 31: Not sure that we need `resources` for the init container, and if so it probably need to have different values than the main container.

charts/apiclarity/templates/deployment.yaml
```diff
@@ -89,12 +91,7 @@ spec:
             allowPrivilegeEscalation: false
             readOnlyRootFilesystem: true
           resources:
-            requests:
-              memory: "200Mi"
-              cpu: "100m"
-            limits:
-              memory: "1000Mi"
-              cpu: "1000m"
+          {{- toYaml .Values.apiclarity.resources | nindent 13 }}
```
FrimIdan, line 94: shouldn't it be `12`?

charts/apiclarity/values.yaml
```diff
@@ -43,6 +43,22 @@ apiclarity:
...
   logLevel: warning
 
+  ## Enable/disable rbac resource creation (i.e. ClusterRole, ClusterRoleBinding)
+  rbac:
+    create: true
+
+  ## Enable/disable ServiceAccount creation
+  serviceAccount:
+    create: true
+  ## Resource limits for clarity deployment
```
FrimIdan, line 53: please add new line before the comment

charts/apiclarity/templates/serviceaccount.yaml
```diff
@@ -1,7 +1,9 @@
+{{- if .Values.apiclarity.serviceAccount.create -}}
```
FrimIdan, line 1: Not sure why it is needed, but if we want to support it, should it be on all SA created in all templates?
Also, what about the `serviceAccount` field in the deployments.

jakparks:
> Hi @jakparks thank you for submitting a PR. Can you please elaborate why `rbac.create` and `serviceAccoutn.create` is needed and how the application will work otherwise? Is it counting on the default service account?

Sure I'll clarify a bit. In our particular k8s environment, application developers don't have RBAC permissions to create their own ClusterRoles and ClusterRoleBindings. Rather, those are provisioned ahead-of-time, in our case there is some automation in place to create them from a git repo that is controlled by the platform team. So we need none to be created at all, we simply bind the Service Account to our pre-existing ClusterRole.

I think this is a pretty common situation/best-practice and you'll often see Helm charts support it. I ripped the general idea from the Kong chart actually, you can see how the support it [here](https://github.com/Kong/charts/blob/fd4cf30c8042a7c4a029df31790a9998f5d69308/charts/kong/templates/controller-rbac-resources.yaml#L1)

charts/apiclarity/templates/deployment.yaml
jakparks: The resources are necessary since it is possible to set a quota which will error if no requests/limits are specified. Fair point though, I'll make it a separate value.

charts/apiclarity/templates/serviceaccount.yaml
jakparks: Good point, I'll make it an option in all templates. Ideally I think I'd like to add an option to override the SA name if this is set to false. It's not necessary to my immediate use-case but it might be a good option to add if you agree...

FrimIdan:
> > Hi @jakparks thank you for submitting a PR. Can you please elaborate why `rbac.create` and `serviceAccoutn.create` is needed and how the application will work otherwise? Is it counting on the default service account?
> 
> Sure I'll clarify a bit. In our particular k8s environment, application developers don't have RBAC permissions to create their own ClusterRoles and ClusterRoleBindings. Rather, those are provisioned ahead-of-time, in our case there is some automation in place to create them from a git repo that is controlled by the platform team. So we need none to be created at all, we simply bind the Service Account to our pre-existing ClusterRole.
> 
> I think this is a pretty common situation/best-practice and you'll often see Helm charts support it. I ripped the general idea from the Kong chart actually, you can see how they support it [here](https://github.com/Kong/charts/blob/fd4cf30c8042a7c4a029df31790a9998f5d69308/charts/kong/templates/controller-rbac-resources.yaml#L1)

Thanks for the clarification, that indeed make sense.

charts/apiclarity/templates/serviceaccount.yaml
FrimIdan: OK, an option to override the SA sounds good. Just make sure all service accounts are covered. Thanks!

jakparks:
Ok, this is ready for review again
- Added serviceAccount.name option. This will override the name used in the deployment, as well as the name on the created SA if serviceAccount.create is true
- Note I didn't touch the names of the ServiceAccounts created for the pre/post-install jobs. This is just the actual service's SA. If we want to add an override for those in the future it should probably be a separate value.

charts/apiclarity/values.yaml
```diff
@@ -43,6 +43,35 @@ apiclarity:
...
+    create: true
+
+  ## ServiceAccount settings
+  serviceAccount:
+    ## Enable/disable ServiceAccount creation. Set false to use a pre-existing account
+    create: true
+    ## Override name of ServiceAccount
+    # name:
+
+  ## Resource limits for clarity deployment
```
FrimIdan, line 57: Resource limits for **APIClarity** deployment

charts/apiclarity/values.yaml
```diff
@@ -43,6 +43,35 @@ apiclarity:
...
+  resources:
+    requests:
+      memory: "200Mi"
+      cpu: "100m"
+    limits:
+      memory: "1000Mi"
+      cpu: "1000m"
+
+  ## Resource limits for init container
```
FrimIdan, line 66: Resource limits for APIClarity init container deployment

FrimIdan (commented):
LGTM @jakparks , just minor spelling changes and we can push it.

FrimIdan (approved):

PR #112, opened 2022-04-06
feat: add the ability to post api inventory items manually

danielvladco:
Add the ability to post api inventory items manually


backend/pkg/rest/apiinventory_controller.go
```diff
@@ -26,6 +26,17 @@ import (
 	_database "github.com/apiclarity/apiclarity/backend/pkg/database"
 )
 
+func (s *Server) PostAPIInventory(params operations.PostAPIInventoryParams) middleware.Responder {
+	apiInfo := &_database.APIInfo{
+		Type: params.Body.APIType,
+		Name: params.Body.Name,
+		Port: params.Body.Port,
+	}
+	s.dbHandler.APIInventoryTable().CreateAPIInfo(apiInfo)
```
FrimIdan, line 35: maybe we should not ignore the error if it's data that we are getting from the user.

FrimIdan: What if that the api already exist, isn't it better to use `FirstOrCreate`? Or to verify that it does not exits (name+port+type) before creating a new one?

danielvladco: I like the idea with FirstOrCreate more, the apis should be idempotent

danielvladco: done

FrimIdan (approved):

FrimIdan (approved):

api/swagger.yaml
```diff
@@ -452,6 +463,21 @@ paths:
                   $ref: '#/definitions/ApiInfo'
         default:
           $ref: '#/responses/UnknownError'
+    post:
+      summary: 'Create API inventory item'
+      parameters:
+        - in: body
+          name: body
+          schema:
+            $ref: '#/definitions/CreateApiInfo'
```
fishkerez, line 472: Why not using ApiInfo?

danielvladco: because we need to provide apiType

fishkerez: then I think it will be better if CreateApiInfo will contain ApiInfo and type instead of duplicating ApiInfo fields.

fishkerez: And I will probably would not call it CreateApiInfo, maybe something more general like ApiInfoWithType that can be reused later?

fishkerez (approved):

akpsgit (approved):

PR #119, opened 2022-04-27
e2e tests

fishkerez:


e2e/utils/utils.go
```diff
@@ -0,0 +1,173 @@
...
+		return fmt.Errorf("failed to load docker image to cluster: %v", err)
+	}
+	if err := LoadDockerImageToCluster(cluster, fmt.Sprintf("ghcr.io/apiclarity/passive-taper:%v", tag)); err != nil {
+		return fmt.Errorf("failed to load docker image to cluster: %v", err)
+	}
+
+	return nil
+}
+
+func LoadDockerImageToCluster(cluster, image string) error {
```
FrimIdan, line 64: why not to use https://github.com/kubernetes-sigs/e2e-framework/blob/main/support/kind/kind.go#L207?

fishkerez: Didn't work for some reason

PR #121, opened 2022-05-03
github: apiclarity to openclarity org move

FrimIdan:


klyr (approved):

charts/apiclarity/Chart.yaml
```diff
@@ -4,9 +4,9 @@ appVersion: latest
 description: Charts for APIClarity deployment. Reconstruct OpenAPI Specifications from real-time workload traffic seamlessly.
```
jadiaconu, line 4: Shall the name APIClarity also change ?

charts/apiclarity/Chart.yaml
```diff
@@ -4,9 +4,9 @@ appVersion: latest
...
 version: latest
 home: https://apiclarity.io/
```
jadiaconu, line 6: Same question for the domain ?

akpsgit (approved):

charts/apiclarity/Chart.yaml
jadiaconu: Ok, so only org changes, ignore this

charts/apiclarity/Chart.yaml
jadiaconu: Ok, so only org changes, ignore this

jadiaconu (approved):

PR #124, opened 2022-05-06
Adds Trace Analyzer module

klyr:


backend/pkg/modules/assets/TraceAnalyzer/sensitive_rules/sensitive_rules_1.yaml
```diff
@@ -0,0 +1,29 @@
+- id: core-001
+  description: Find 'password' keyword in flow data
+  regex: '([pP][aA][sS][sS][wW][oO][rR][dD])'
```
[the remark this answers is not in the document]
klyr: The rules are regexps written by the user. It's not means to be generated.
This file is an example.

At the moment, list of sensitive keywords are only checked inside the JWT claims and headers.

backend/pkg/modules/assets/TraceAnalyzer/sensitive_keywords/tests.txt
```diff
@@ -0,0 +1,4 @@
+password
```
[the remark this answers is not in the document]
klyr: This is an example file of sensitive keywords that can be seen in JWT claims and headers.

I replaced it with https://raw.githubusercontent.com/danielmiessler/SecLists/master/Discovery/Variables/secret-keywords.txt

backend/pkg/modules/internal/traceanalyzer/guessableid/guessableid.go
```diff
@@ -0,0 +1,240 @@
...
+			if err != nil {
+				sMatrix[i][j] = 0.0
+			} else {
+				sMatrix[i][j] = res
+			}
+		}
+	}
+
+	// Frobenius Norm
+	var sum float32
```
danielvladco, line 184: We could use some preexisting math functions so we don't reinvent the wheel

Ex: https://pkg.go.dev/gonum.org/v1/gonum/mat#Norm

Although in Go it's sometimes recommended to use a simple function instead of importing a new package or adding an entire module I think we should use gonum. Since we will be adding gonum anyway for the stats modules and it will also make the code simpler.

If you are still not convinced, at least separate this math in a different function

backend/pkg/modules/internal/traceanalyzer/nlid/nlid.go
```diff
@@ -0,0 +1,280 @@
...
+const (
+	MinIDValueLength = 8
+	MaxIDValueLength = 40 // This is at least a UUID. every string larger that than is considered low chance of being an ID
+)
+
+var IDKeys = [...]string{"id", "ids", "identifier", "identifiers"}
+
+type params = map[string]bool
+
+type Reason map[string]interface{}
```
danielvladco, line 47: I don't see where it is being used

backend/pkg/modules/internal/traceanalyzer/nlid/nlid.go
```diff
@@ -0,0 +1,280 @@
...
+		// Nothing, it's not interresting
+	case json.Number:
+		// We only add integers to the possible list of identifier.
+		// Floats are unlikely indentifiers.
+		if n, err := val.Int64(); err == nil {
+			s := fmt.Sprintf("%d", n)
+			if maybeID(prefix, s) {
+				(*o)[s] = true
+			}
+			// } else if _, err := val.Float64(); err == nil {
```
danielvladco, line 222: dead code

backend/pkg/modules/internal/traceanalyzer/nlid/nlid.go
```diff
@@ -0,0 +1,280 @@
...
+	if err := d.Decode(&parsed); err != nil {
+		return fmt.Errorf("unable to decode body: %w", err)
+	}
+
+	getDecodedBodyParams(params, parsed, "")
+
+	return nil
+}
+
+func getDecodedBodyParams(o *map[string]bool, val interface{}, prefix string) {
```
danielvladco, line 210:
```suggestion
func getDecodedBodyParams(o map[string]bool, val interface{}, prefix string) {
```
No need for passing map as a pointer since map is already a reference

backend/pkg/modules/internal/traceanalyzer/guessableid/guessableid.go
klyr: The Frobenius norm is the commented part.
This is not relevant to use a separate library to just do the sum if the elements of the matrix.
Maybe we will go back to that in the future.

backend/pkg/modules/internal/traceanalyzer/nlid/nlid.go
klyr: it is on https://github.com/openclarity/apiclarity/blob/3f7fb314332913541f57c1185ac973ab2aecff7c/backend/pkg/modules/internal/traceanalyzer/nlid/nlid.go#L252

klyr: It's used in traceanalyzer.go

danielvladco (approved):

PR #131, opened 2022-06-13
OpenAPIv3 support

FrimIdan:
pending https://github.com/openclarity/speculator/pull/20


backend/go.mod
```diff
@@ -138,3 +136,5 @@ replace github.com/openclarity/trace-sampling-manager/api v0.0.0 => github.com/o
 replace github.com/openclarity/apiclarity/api v0.0.0 => ./../api
 
 replace github.com/openclarity/apiclarity/plugins/api v0.0.0 => ./../plugins/api
+
+//replace github.com/openclarity/speculator v0.0.0 => /Users/idanf/gocode/src/github.com/openclarity/speculator
```
akpsgit, line 140: remove

backend/pkg/backend/backend.go
```diff
@@ -220,9 +221,33 @@ func convertSpecDiffToEventDiff(diff *_spec.APIDiff) (originalRet, modifiedRet [
...
+	if v3PathItem == nil {
+		return v3PathItem
+	}
+
+	switch version {
+	case _spec.OASv2:
+		log.Errorf("Converting to OASv2 path item")
+		v2PathItem, err := openapi2conv.FromV3PathItem(&spec.T{Components: spec.Components{}}, v3PathItem)
+		if err != nil {
+			log.Errorf("failed to convert v3 path item to v2, keeping v3: %v", err)
```
akpsgit, line 234: Failed to convert.
Don't we want to return an error in this case (the function can't do what the user requested)?

backend/pkg/backend/backend_test.go
```diff
@@ -750,3 +756,116 @@ func TestBackend_handleHTTPTrace(t *testing.T) {
...
+		v3PathItem *spec.PathItem
+		version    _spec.OASVersion
+	}
+	tests := []struct {
+		name string
+		args args
+		want interface{}
+	}{
+		{
+			name: "v2",
```
akpsgit, line 771: we can add the nil test

backend/pkg/backend/backend.go
FrimIdan: would it not be better to show diff in v3 format rather than not showing anything?

backend/pkg/rest/apiinventory_get_spec_json_controller.go
```diff
@@ -63,28 +66,49 @@ func (s *Server) GetAPIProvidedSwaggerJSON(params operations.GetAPIInventoryAPII
...
-	if specToReturn == "" {
+	if len(specToReturn) == 0 {
 		return nil, fmt.Errorf("%v spec not found", typ)
 	}
 
-	analyzed, err := loads.Analyzed([]byte(specToReturn), "")
+	specToReturn, err = yaml.YAMLToJSON(specToReturn)
 	if err != nil {
-		return nil, fmt.Errorf("failed to analyzed spec: %v", err)
+		return nil, fmt.Errorf("failed to convert provided spec into json: %s. %v", specToReturn, err)
```
akpsgit, line 89: provided? :)

backend/pkg/backend/backend.go
akpsgit: we do, but it's a bit misleading. maybe better to call it getPathItemForVersionOrOriginal?

akpsgit (approved):

PR #140, opened 2022-06-20
Aduminuc api3 docs

aduminuc:
Fail pipeline in case the api spec aggregation and code generation was not run on the latest version of specs

### DIFF
--- a/api3/generate.sh
+++ b/api3/generate.sh
@@ -1,0 +1,16 @@
+#!/bin/bash	
+
+
+(cd ../tools/spec-aggregator && go run main.go)
+(cd ./common && go generate)
+(cd ./global && go generate)
+(cd ./notifications && go generate)
+
+if [[ $1 == "--verify" ]]  ; then
+    diffs=$(git status --porcelain)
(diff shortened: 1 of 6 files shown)


tools/spec-aggregator/main.go
```diff
@@ -205,6 +206,11 @@ func aggregateNotificationSpecs() {
 			}
 		}
 	}
+	a := parentNotificationSchema.OneOf[0].Ref
+	_ = a
```
FrimIdan, line 210: what is this for?

aduminuc: Leftover of a test. Cleaning up.

FrimIdan (approved):

klyr (approved):
LGTM

api3/generate.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash	
+
+(cd ../tools/spec-aggregator && go run main.go)
+(cd ./common && go generate)
+(cd ./global && go generate)
+(cd ./notifications && go generate)
```
klyr, line 6: go generate ./...

PR #145, opened 2022-06-21
Man 9496 new report

axelcisco:
Stuff for MAN-9496: 
- Add a new report
- Support common/global API openapi spec

### DIFF
--- a/backend/go.mod
+++ b/backend/go.mod
@@ -67,6 +67,7 @@ require (
 	github.com/google/gofuzz v1.2.0 // indirect
 	github.com/google/uuid v1.3.0 // indirect
 	github.com/googleapis/gnostic v0.5.5 // indirect
+	github.com/gorilla/mux v1.8.0 // indirect
 	github.com/hashicorp/hcl v1.0.0 // indirect
 	github.com/imdario/mergo v0.3.12 // indirect
 	github.com/invopop/yaml v0.1.0 // indirect
@@ -133,15 +134,17 @@ require (
 	google.golang.org/protobuf v1.28.0 // indirect
 	gopkg.in/inf.v0 v0.9.1 // indirect
 	gopkg.in/ini.v1 v1.62.0 // indirect
-	gopkg.in/yaml.v2 v2.4.0 // indirect
 	k8s.io/klog/v2 v2.30.0 // indirect
 	k8s.io/kube-openapi v0.0.0-20211115234752-e816edb12b65 // indirect
-	k8s.io/utils v0.0.0-20211116205334-6203023598ed // indirect
 	sigs.k8s.io/json v0.0.0-20211020170558-c049b76a60c6 // indirect
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.1 // indirect
 )
 
-require github.com/openclarity/apiclarity/api3 v0.0.0
+require (
+	github.com/openclarity/apiclarity/api3 v0.0.0
+	gopkg.in/yaml.v2 v2.4.0
+	k8s.io/utils v0.0.0-20211116205334-6203023598ed
+)
 
 replace github.com/openclarity/trace-sampling-manager/api v0.0.0 => github.com/openclarity/trace-sampling-manager/api v0.0.0-20220503091058-2730f5ebad2c
 
--- a/backend/pkg/modules/internal/fuzzer/clients/fake.go
+++ b/backend/pkg/modules/internal/fuzzer/clients/fake.go
@@ -79,8 +79,10 @@ func FakeTriggerFuzzingJob(ctx context.Context, testFilename string, apiID uint,
 		logging.Logf("[Fuzzer] Use data from (%v)", testFilename)
 		scanner := bufio.NewScanner(file)
 		// optionally, resize scanner's capacity for lines over 64K
+		buf := make([]byte, 0, 64*1024)
+		scanner.Buffer(buf, 10*1024*1024)
 		for scanner.Scan() {
-			logging.Logf("[Fuzzer] inject (%v)", scanner.Text())
+			logging.Logf("[Fuzzer] inject data len=(%v)", len(scanner.Text()))
 			err = SendReport(ctx, remoteHost, apiID, scanner.Bytes()) // TODO handle error
 			if err != nil {
 				logging.Errorf("Failed to send report to (%v): %v", remoteHost, err)
--- a/backend/pkg/modules/internal/fuzzer/tools/specs.go
+++ b/backend/pkg/modules/internal/fuzzer/tools/specs.go
@@ -4,24 +4,46 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net/http"
+
+	"github.com/getkin/kin-openapi/openapi2"
+	"github.com/getkin/kin-openapi/openapi2conv"
+	"github.com/getkin/kin-openapi/openapi3"
+	"github.com/getkin/kin-openapi/routers"
+	"k8s.io/utils/strings/slices"
 
 	"github.com/openclarity/apiclarity/api/server/models"
 	"github.com/openclarity/apiclarity/backend/pkg/modules/internal/core"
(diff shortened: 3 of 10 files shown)


aduminuc (approved):
LGTM

backend/pkg/modules/internal/fuzzer/tools/specs.go
```diff
@@ -1,36 +1,163 @@
...
+	return specV2OrV3.Openapi != nil
+}
+
+func LoadSpec(spec []byte) (*openapi3.T, error) {
+	if IsV2Specification(spec) {
+		logging.Logf("[Fuzzer] getDocFromSpec(): spec V2 identified")
+		loader := openapi3.NewLoader()
+		var docV2 openapi2.T
+		if err := json.Unmarshal(spec, &docV2); err != nil {
+			panic(err)
```
FrimIdan, line 107: better not to use `panic` and return the error back for better error handling.
(true for all files)

backend/pkg/modules/internal/fuzzer/tools/specs.go
```diff
@@ -1,36 +1,163 @@
...
+		if err := doc.Validate(loader.Context); err != nil {
+			panic(err)
+		}
+		return doc, nil
 	}
+	return nil, fmt.Errorf("invalid spec")
+}
 
-	return specsInfo, nil
+func GetBasePathsFromServers(servers *openapi3.Servers) []string {
```
FrimIdan, line 132: I've implemented the same differently [here](https://github.com/openclarity/speculator/pull/20/files#diff-15c06e366d2e41835cf86a9108e014f70974fb443ac61910273273b1e855f7faR103) to support servers w/o host names and with formatting (examples can be found [here](https://swagger.io/docs/specification/api-host-and-base-path/) and in the [unit test](https://github.com/openclarity/speculator/pull/20/files#diff-93600c0940ef11d86724a8e7a2b8df94c11df899d005e6a2067bdd1fb0e9e0c2R204))
maybe we can use import the logic from the speculator once merged. WDYT?

backend/pkg/modules/internal/fuzzer/tools/specs.go
```diff
@@ -1,36 +1,163 @@
...
+	//nolint:noctx // No need of context, the http.NewRequest is used only for formatting
+	req, err := http.NewRequest(verb, uri, nil)
+	if err != nil {
+		return nil, fmt.Errorf("can't convert (%v %v) to http request, err=(%v)", verb, uri, err.Error())
+	}
+	req.Header.Set("Content-Type", "application/json") // Report this path in shortreport
+	logging.Debugf("[Fuzzer] findRoute(): ... req to find (%v %v)", verb, req)
+	route, _, err := (*router).FindRoute(req)
+	if err != nil {
+		return nil, fmt.Errorf("can't find route for (%v %v), err=(%v)", verb, uri, err.Error())
```
FrimIdan, line 160: `err.Error()` is not needed. you can just use `err`

backend/pkg/modules/internal/fuzzer/tools/tools.go
```diff
@@ -120,3 +135,15 @@ func DumpHTTPFuzzParam(params restapi.FuzzTargetParams) string {
 	ret = ret + "}"
 	return ret
 }
+
+//nolint:varnamelen
+func TrimLeftChars(s string, n int) string {
```
FrimIdan, line 140: please add unit test.
also, not sure you need to loop over s, why not just do
```
	if n > len(s) {
		return ""
	}
	return s[n:]
```

backend/pkg/modules/internal/fuzzer/tools/specs.go
```diff
@@ -1,36 +1,163 @@
...
+
+	return &fuzzerSpecsInfo, nil
+}
+
+func IsV2Specification(data []byte) bool {
+	specV2OrV3 := struct {
+		Openapi *string `json:"openapi"`
+		Swagger *string `json:"swagger"`
+	}{}
+	if err := yaml.Unmarshal(data, &specV2OrV3); err != nil {
```
FrimIdan, line 84: will it work if the input is json?

backend/go.mod
```diff
@@ -133,15 +134,17 @@ require (
...
 	sigs.k8s.io/json v0.0.0-20211020170558-c049b76a60c6 // indirect
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.1 // indirect
 )
 
-require github.com/openclarity/apiclarity/api3 v0.0.0
+require (
+	github.com/openclarity/apiclarity/api3 v0.0.0
+	gopkg.in/yaml.v2 v2.4.0
+	k8s.io/utils v0.0.0-20211116205334-6203023598ed
+)
```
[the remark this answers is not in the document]
axelcisco: ok

backend/pkg/modules/internal/fuzzer/clients/fake.go
```diff
@@ -79,8 +94,10 @@ func FakeTriggerFuzzingJob(ctx context.Context, testFilename string, apiID uint,
 		logging.Logf("[Fuzzer] Use data from (%v)", testFilename)
 		scanner := bufio.NewScanner(file)
 		// optionally, resize scanner's capacity for lines over 64K
+		buf := make([]byte, 0, 64*1024)
+		scanner.Buffer(buf, 10*1024*1024)
```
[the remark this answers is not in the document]
axelcisco: ok

backend/pkg/modules/internal/fuzzer/tools/specs.go
axelcisco: You totally right. I use 'panic' only at first stage of implementation, don't know why I skip this file when I took them off. 
Done.

backend/pkg/modules/internal/fuzzer/tools/specs.go
axelcisco: If I'm right, you can specify one or more base URLs on openapi specs on "servers" part, then we need to return an array of string, at least for Fuzzer needs.
For the implementation, sure your code works. Using http.NewRequest in place of "manual" parsing allow to manage all "strange use cases" we can encountered when a user fill a URL manually (and use less of code).
But it doesn't matter, if you prefer I can reuse your code. But I will have to change it to manage list of string, and then I will have to change the behavior in all ApiClarity codes that use your function...

backend/pkg/modules/internal/fuzzer/tools/specs.go
axelcisco: ok

backend/pkg/modules/internal/fuzzer/tools/specs.go
axelcisco: yes. yaml.unmarshal support json AND yaml (in fact, in case of yaml, it does the yaml->json conversion)

backend/pkg/modules/internal/fuzzer/tools/tools.go
axelcisco: It was an implementation found somewhere. But following your remarks I see that "strings.TrimPrefix" exists and do exactly what I need. I will remove this "trimLeftChars" function.
Thanks.

backend/pkg/modules/internal/fuzzer/tools/specs.go
FrimIdan: ok, I understand. does `http.NewRequest` handle templating, for example `{protocol}://api.example.com/api`?

backend/pkg/modules/internal/fuzzer/tools/specs.go
FrimIdan: not sure why, interesting.

backend/pkg/modules/internal/fuzzer/tools/specs.go
axelcisco: No, you're right. I was not aware about templating in servers entries. I tested it (in case of) but of course it doesn't work. 
I will fix this.

PR #147, opened 2022-06-21
cleanup of spec-aggregator tool

aduminuc:


tools/spec-aggregator/main.go
```diff
@@ -12,28 +13,32 @@ import (
...
+)
 
 func loadSpec(path string, fail bool) *openapi3.T {
 	loader := openapi3.NewLoader()
 	loader.IsExternalRefsAllowed = true
 
 	spec, err := loader.LoadFromFile(path)
 	if err != nil {
-		log.Errorf("Unable to load spec %s: %v", path, err)
+		log.Warnf("Unable to load spec %s: %v", path, err)
```
FrimIdan, line 35: if we don't want to fail so no need to print it as a warning. we can just put the message in the `log.Fatal`

tools/spec-aggregator/main.go
```diff
@@ -73,131 +81,166 @@ func aggregateGlobalSpecs() {
...
-			coreSpec.Paths["/modules/"+module+k] = v
+			coreSpec.Paths["/modules/"+module+key] = value
 		}
 	}
+
+	/* now coreSpec holds the new global spec. We can write to file */
 	specout, err := coreSpec.MarshalJSON()
 	if err != nil {
-		log.Errorf("ERROR: %v", err)
+		log.Fatalf("ERROR: %v", err)
```
FrimIdan, line 128: `ERROR:` is not needed, true for the entire file

FrimIdan (approved):

PR #155, opened 2022-06-23
make highest severity optional in fuzzer reports

aduminuc:


api3/global/openapi.gen.yaml
```diff
@@ -603,7 +603,6 @@ components:
       - operation
       - requestsCount
       - findings
-      - highestSeverity
```
FrimIdan, line 606: I think it should be required in  `FuzzingReportOperation` unless `findings` can be empty (and then `findings` should not be required but I don't think this is the case here)

api3/global/openapi.gen.yaml
```diff
@@ -813,7 +811,6 @@ components:
       - starttime
       - status
       - tags
```
FrimIdan, line 813: what about tags? should it be `required`? in case of and error tags should be empty I think.

api3/global/openapi.gen.yaml
```diff
@@ -638,7 +637,6 @@ components:
       required:
       - name
       - operations
-      - highestSeverity
```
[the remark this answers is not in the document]
aduminuc: You can have operations with no findings, where only the requestCount is populated, in that case highestSeverity would be empty. We can leave operations as mandatory, if we really end up in a case (which I don't expect at the moment) where there are no operations, then the array can be empty

api3/global/openapi.gen.yaml
aduminuc: We can have no findings, so I'll make it optional.

axelcisco (approved):
LGTM

FrimIdan (approved):

PR #158, opened 2022-06-27
[fuzzer] feat: start function with depth parameter

axelcisco:
- Implement new StartTest() operation
- Implement new start test return format.


klyr (approved):
LGTM

PR #159, opened 2022-06-27
Diffs notificaition swagger

fishkerez:


api3/notifications/openapi.gen.yaml
```diff
@@ -104,6 +106,19 @@ components:
       - highestSeverity
       title: Short Test Report
       type: object
+    SpecDiffs:
```
FrimIdan, line 109: seems like `SpecDiffs` should be defined in the common swagger since both global and notifications uses it

api3/global/openapi.gen.yaml
```diff
@@ -526,6 +526,37 @@ components:
       - id
       - source
       - ip_address
+    DiffType:
```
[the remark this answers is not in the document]
fishkerez: This is how it generated... I only defined the swagger in **backend/pkg/modules/internal/differ/restapi/openapi.yaml**

api3/notifications/openapi.gen.yaml
fishkerez: This is how it generated... I only defined the swagger in backend/pkg/modules/internal/differ/restapi/openapi.yaml

FrimIdan (approved):

PR #163, opened 2022-07-01
[Fuzzer] feat: add fuzzing job cancellation

axelcisco:
feat: add cancellation of Fuzzing job on Module Fuzzer
fix: improper value for additionalInfo on Module Fuzzer report


backend/pkg/modules/internal/fuzzer/clients/kubernetes.go
```diff
@@ -98,6 +99,24 @@ func (l *K8sClient) TriggerFuzzingJob(apiID int64, endpoint string, securityItem
...
+		return fmt.Errorf("no current k8s job to terminate")
+	}
+	var zero int64 // = 0
+	policy := metav1.DeletePropagationForeground
+	deleteOptions := &metav1.DeleteOptions{
+		GracePeriodSeconds: &zero,
+		PropagationPolicy:  &policy,
+	}
+	err := l.hClient.BatchV1().Jobs(l.currentJob.Namespace).Delete(context.TODO(), l.currentJob.Name, *deleteOptions)
+	l.currentJob = nil
```
FrimIdan, line 113: why clearing `currentJob` here and not after error check?

backend/pkg/modules/internal/fuzzer/clients/kubernetes.go
```diff
@@ -98,6 +99,24 @@ func (l *K8sClient) TriggerFuzzingJob(apiID int64, endpoint string, securityItem
...
+	if err != nil {
+		return fmt.Errorf("failed to stop k8s fuzzer job")
```
FrimIdan, line 115: please also return the error message
```
return fmt.Errorf("failed to stop k8s fuzzer job: %v", err)
```

backend/pkg/modules/internal/fuzzer/clients/kubernetes.go
axelcisco: To reset l.currentJob in all case. even if error. But I can put it after if you want, I don't test it for now before the start (if a such test exists, we will have a lock for ever in case of error during the stop)

backend/pkg/modules/internal/fuzzer/clients/kubernetes.go
axelcisco: ok

aduminuc (approved):

backend/pkg/modules/internal/fuzzer/fuzzer.go
```diff
@@ -235,6 +239,113 @@ func (p *pluginFuzzer) FuzzTarget(ctx context.Context, apiID oapicommon.ApiID, p
...
+	api, err := p.model.GetAPI(ctx, uint(apiID))
+	if err != nil {
+		// Must not happen, as we had be able to start a Fuzzing...
+		logging.Errorf("[Fuzzer] StopFuzzing(): can't retrieve API (%v)", apiID)
+		return &NotFoundError{msg: ""}
+	}
+
+	// Some checks...
+	if p.fuzzerClient == nil {
+		// Must not happen, as we had be able to start a Fuzzing...
```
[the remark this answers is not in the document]
axelcisco: done

PR #165, opened 2022-07-02
Aduminuc selective traces

aduminuc:
- Per module selective tracing
- aggregation of feature enum in specs
- Still miss the selective tracing for spec diffs and spec reconstruction


backend/pkg/modules/internal/traceanalyzer/traceanalyzer.go
```diff
@@ -548,6 +548,39 @@ func (h httpHandler) GetApiFindings(w http.ResponseWriter, r *http.Request, apiI
 	json.NewEncoder(w).Encode(apiFindings)
 }
 
+func (h httpHandler) StartTraceAnalyis(w http.ResponseWriter, r *http.Request, apiID oapicommon.ApiID) {
```
klyr, line 551: Typo: s/TraceAnalyis/TraceAnalysis/g, in all files

api/swagger.yaml
```diff
@@ -331,7 +331,7 @@ definitions:
       - ALERT_INFO
       - ALERT_WARN
       - ALERT_CRITICAL
-
+
```
klyr, line 334:
```suggestion
```

api3/core/openapi.yaml
```diff
@@ -772,6 +783,42 @@ components:
...
+      - featureName
+
+    APIClarityFeatureEnum:
+      description: 'APIClarity Feature Name'
+      type: string
+      enum:
+        - specreconstructor
+        - specdiffs
+        
+    APIClarityFeatureOperationEnum:
```
akpsgit, line 816: this is not used

api3/core/openapi.yaml
```diff
@@ -772,6 +783,42 @@ components:
...
+            $ref: '#/components/schemas/APIClarityFeature'
+    APIClarityFeature:
+      type: 'object'
+      properties:
+        featureName:
+          $ref: '#/components/schemas/APIClarityFeatureEnum'
+        featureDescription:
+          description: "Short human readable description of the feature"
+          type: string
+        hostsToTrace:
```
akpsgit, line 802: we should add a description to this. e.g. "List of hosts (with ports) that the feature requires to get trace for"

api/swagger.yaml
```diff
@@ -547,6 +586,15 @@ paths:
               $ref: '#/definitions/SpecDiffTime'
         default:
           $ref: '#/responses/UnknownError'
+          
+  /features:
+    get:
+      summary: 'Get the list of APIClarity features'
```
akpsgit, line 592: I saw that there is an unused feature operation enum (APIClarityFeatureOperationEnum). is this list of enabled (added) features, or list of all features known to the system?
Should this be "Get the list of APIClarity active features"?

backend/pkg/backend/backend.go
```diff
@@ -85,6 +86,16 @@ func createDatabaseConfig(config *_config.Config) *_database.DBConfig {
 
 const defaultChanSize = 100
 
+func getCoreFeatures() []modules.ModuleInfo {
+	features := []modules.ModuleInfo{
+		{
+			Name:        "specreconstructor",
```
akpsgit, line 92: please use APIClarityFeatureEnumSpecreconstructor

backend/pkg/backend/backend.go
```diff
@@ -57,17 +58,17 @@ type Backend struct {
 	monitor             *k8smonitor.Monitor
 	apiInventoryLock    sync.RWMutex
 	dbHandler           _database.Database
-	modules             modules.Module
+	modulesWrapper      modules.Module
```
akpsgit, line 61: a bit confused, modules.Module isn't the interface for a single module implementation? why is a modules wrapper?

api/swagger.yaml
aduminuc: This is the list of features known by the system , so for a given running instance of APIClarity the list is always the same.

OperationEnum is a left over of the PUT operation and I am removing it

backend/pkg/backend/backend.go
aduminuc: The module and the modulesWrapper share the same interface. backend calls functions on the wrapper which in turn dispatches them to each of the modules.

api3/core/openapi.yaml
aduminuc: leftover. Removing it.

backend/pkg/backend/backend.go
aduminuc: Done

backend/pkg/modules/internal/bfla/bfladetector/learn_and_detect_bfla_test.go
```diff
@@ -115,9 +115,9 @@ func initBFLADetector(ctrl *gomock.Controller, backendAccessor *core.MockBackend
 	}).AnyTimes()
 	backendAccessor.EXPECT().CreateAPIEventAnnotations(ctx, gomock.Any(), gomock.Any(), gomock.Any()).AnyTimes()
 	backendAccessor.EXPECT().GetAPIInfoAnnotation(ctx, gomock.Any(), gomock.Any(), gomock.Any()).AnyTimes()
-	ctrlNotifier := bfladetector.NewControllerNotifier(backendAccessor)
+	ctrlNotifier := bfladetector.NewControllerNotifier("bfla", backendAccessor)
```
akpsgit, line 118: Several questions:
1. why the method called ControllerNotifier?
2. maybe it's better to have the names returned statically inside the modules in the Name() method?
3. should these names align to the generated API enums?

tools/spec-aggregator/main.go
```diff
@@ -85,11 +86,15 @@ func aggregateGlobalSpecs() {
 	/* loading corespec and then enriching it with the module spec elements */
 	coreSpec := loadSpec(rootPath+"/"+coreSpecPath, true)
 
+	featureEnum := coreSpec.Components.Schemas[featureEnumName]
+	_ = featureEnum
```
akpsgit, line 90: featureEnum is not used right (this was needed to suppress the warning)?

backend/pkg/modules/internal/core/modules.go
```diff
@@ -78,13 +86,14 @@ func NewAccessor(dbHandler *database.Handler, clientset kubernetes.Interface) Ba
 		n.Start(context.Background())
 	}
 
-	return &accessor{dbHandler, clientset, n}
+	return &accessor{dbHandler, clientset, n, samplingManager}
```
akpsgit, line 89: better:
```
	return &accessor{
		dbHandler: dbHandler,
		clientset: clientset,
		notifier: n,
		samplingManager: samplingManager,
	}
```

aduminuc: ack. Fixing it.

backend/pkg/modules/internal/bfla/bfladetector/learn_and_detect_bfla_test.go
akpsgit: sounds good :)

api3/core/openapi.yaml
akpsgit: its host:port
I think that we can put a description similar to the trace manager api:
https://github.com/openclarity/trace-sampling-manager/blob/main/api/swagger.yaml#L36

backend/pkg/backend/backend.go
akpsgit: I see that the same interface can be used, but i think that the code is not very readable this way, because when we go to the interface description, it's says that it's an interface for a specific module. Maybe we shouldn't abstract it and call the wrapper struct directly (maybe worth to rename it to moduleManager or something), or create a separate interface if we need mocks for testing. WDYT?

aduminuc: I'll create another interface that refers to this one, so it will be clear

api3/core/openapi.yaml
aduminuc: I took a look at the envoy plugin code https://github.com/openclarity/wasm-filters/blob/0ae3f8fe2e4e89068c617d145b6086651ed48ac2/src/trace/main.go and I've found this. If I read it correctly this assumes that the hostsToTrace contain just the host part and not host:port. Is it correct?

`func (ctx *TraceFilterContext) shouldTrace() bool {
	if !ctx.isHostFixed {
		return true
	}
	if !ctx.enableTraceSampling {
		return true
	}

	_, foundInApiToTrace := ctx.hostsToTrace[ctx.Telemetry.Request.Host]
	// check if we should trace all hosts
	_, shouldTraceAll := ctx.hostsToTrace["*"]

	if !foundInApiToTrace && !shouldTraceAll {
		proxywasm.LogDebugf("Host should not be traced. host=%v, foundInApiToTrace=%v, shouldTraceAll=%v",
			ctx.Telemetry.Request.Host, foundInApiToTrace, shouldTraceAll)
		return false
	}
	proxywasm.LogDebugf("Host should be traced. host=%v, foundInApiToTrace=%v, shouldTraceAll=%v",
		ctx.Telemetry.Request.Host, foundInApiToTrace, shouldTraceAll)
	return true
}`

aduminuc: OK, making the needed changes to handle host:port

akpsgit (approved):

PR #153, opened 2022-06-23
upgrade bitnami postgres chart

esara:

PR #169, opened 2022-07-05
Charts/update postgresql

pbalogh-sa:
Duplicates: #153


charts/apiclarity/values.yaml
```diff
@@ -90,21 +90,26 @@ apiclarity:
...
-  ## Secret with scripts to be run at first boot (in case it contains sensitive information)
-  ## NOTE: This can work along initdbScripts or initdbScriptsConfigMap
-  # initdbScriptsSecret:
-
-  ## Specify the PostgreSQL username and password to execute the initdb scripts
-  # initdbUser:
-  # initdbPassword:
+  image:
+  ## Use specified image tag
+    # tag: 14.4.0-debian-11-r4
```
FrimIdan, line 95: maybe better to keep it uncomment?

pbalogh-sa: The image should be commented out as well.

FrimIdan (approved):

PR #171, opened 2022-07-06
Fix/add the tags to spec

danielvladco:


aduminuc (commented):
I understand you are doing 2 things in this PR:
1) Adding the information of the tag to the auth model
2) Make sure that also the Aprrove/Deny/Update operations can return error through the channel. 
Is it correct?

backend/pkg/modules/internal/bfla/bfladetector/learn_and_detect_bfla.go
```diff
@@ -88,9 +88,9 @@ type BFLADetector interface {
 	ApproveTrace(path, method string, clientRef *k8straceannotator.K8sObjectRef, apiID uint, user *DetectedUser)
 	DenyTrace(path, method string, clientRef *k8straceannotator.K8sObjectRef, apiID uint, user *DetectedUser)
 
-	ResetLearning(apiID uint, numberOfTraces int)
-	StartLearning(apiID uint, numberOfTraces int)
-	StopLearning(apiID uint)
+	ResetLearning(apiID uint, numberOfTraces int) error
+	StartLearning(apiID uint, numberOfTraces int) error
+	StopLearning(apiID uint) error
```
aduminuc, line 93: Not sure why these appear as differences, in the devel-global-specs I already see them with the "error" return.

https://github.com/openclarity/apiclarity/blob/devel-global-specs/backend/pkg/modules/internal/bfla/bfladetector/learn_and_detect_bfla.go

backend/pkg/modules/internal/bfla/bfladetector/learn_and_detect_bfla.go
```diff
@@ -88,9 +88,9 @@ type BFLADetector interface {
 	ApproveTrace(path, method string, clientRef *k8straceannotator.K8sObjectRef, apiID uint, user *DetectedUser)
 	DenyTrace(path, method string, clientRef *k8straceannotator.K8sObjectRef, apiID uint, user *DetectedUser)
```
aduminuc, line 89: shall we return the error also for Approve/Deny/Update?

danielvladco: let's keep this in a separate discussion

aduminuc (approved):
LGTM

PR #182, opened 2022-07-12
Update helm chart

klyr:


charts/apiclarity/values.yaml
```diff
@@ -39,6 +39,14 @@ global:
 
 apiclarity:
 
+  # Send notification of new API Findings to this backend
+  # notification_url_prefix: http://example.local:8888/
```
FrimIdan, line 43: Please use camel case `notificationUrlPrefix`
Also, `Send notification of new API Findings to this backend` - it is not only for `new API Findings`, right?

FrimIdan (approved):

PR #191, opened 2022-07-18
traceanalyzer: Remove dead code

klyr:


backend/pkg/modules/internal/traceanalyzer/sensitive/findings.go
```diff
@@ -103,15 +103,6 @@ func (a APIAnnotationRegexpMatching) Redacted() utils.TraceAnalyzerAPIAnnotation
 	return &newA
 }
 
-func (a *APIAnnotationRegexpMatching) ToFinding() utils.Finding {
-	return utils.Finding{
-		ShortDesc:    "Matching regular expression",
```
aduminuc, line 108: Is this your internal format? So Short Desc==Name?

klyr: It was the previous format, served for the UI.
Now, we are using APIFindings in the UI.

PR #193, opened 2022-07-18
Adds license cache

klyr:


aduminuc (commented):
Why do we need this?

FrimIdan:
> I don't understand if it's needed, sometimes the licence check complains, but most of the time it does not, do you understand why?

@aduminuc / @klyr 
So the reason that license check complains sometimes is due to GitHub rate limit. As I understand it license check used GH token to fetch repositories and look for license file. GH sometimes blocks those fetch actions even when using access token. (https://github.com/goph/licensei#github-authentication)
It make sense to maintain a cache to avoid repetitive fetch calls to the same repo over and over, i didn't had time to research it more.

klyr:
Resurrect this PR.

aduminuc (approved):

PR #198, opened 2022-07-20
fix: bfla escape location with slashes

danielvladco:


klyr (approved):
We'll move this function to a common one, not only in the trace analyzer

PR #200, opened 2022-07-20
Move JSONPointer function to common location

klyr:


aduminuc (approved):
LGTM, we need to ask Daniel and Axel to use this.

PR #206, opened 2022-07-24
feat: TLS support

FrimIdan:


backend/pkg/rest/server.go
```diff
@@ -149,13 +163,21 @@ func CreateRESTServer(port int, speculator *_speculator.Speculator, dbHandler *d
 
 	server.ConfigureFlags()
 	server.ConfigureAPI()
-	server.Port = port
+	server.Port = config.Port
+
+	// We want to serve both http and https
+	if config.EnableTLS {
+		server.EnabledListeners = []string{"https", "http"}
+		server.TLSCertificate = flags.Filename(config.TLSServerCertFilePath)
```
fishkerez, line 171: maybe we want to check that the certs actually exists?

charts/apiclarity/templates/deployment.yaml
```diff
@@ -79,8 +89,13 @@ spec:
               value: kubernetes
             - name: FUZZER_IMAGE_NAME
               value: {{ .Values.apiclarity.modules.fuzzer.docker.image }}
+            {{- if .Values.apiclarity.tls.enabled }}
+            - name: FUZZER_PLATFORM_HOST_FROM_FUZZER
+              value: http://{{ include "apiclarity.name" . }}.{{ .Release.Namespace }}:8080/api
```
fishkerez, line 94: if else same values, should it be https?

plugins/gateway/tyk/deploy/deploy.sh
```diff
@@ -2,12 +2,43 @@
 TykProxyContainerName="${TYK_PROXY_CONTAINER_NAME:-tyk-gtw}"
 TykGatewayDeploymentName="${TYK_GATEWAY_DEPLOYMENT_NAME:-tyk-gtw}"
 TykGatewayDeploymentNamespace="${TYK_GATEWAY_DEPLOYMENT_NAMESPACE:-default}"
-UpstreamTelemetryHostName="${UPSTREAM_TELEMETRY_HOST_NAME:-apiclarity-apiclarity.apiclarity:9000}"
+UpstreamTelemetryHostName="${UPSTREAM_TELEMETRY_HOST_NAME:-apiclarity-apiclarity.apiclarity}"
+UpstreamTelemetryHTTPPort="${UPSTREAM_TELEMETRY_HTTP_PORT:-9000}"
+UpstreamTelemetryTLSPort="${UPSTREAM_TELEMETRY_TLS_PORT:-9443}"
```
fishkerez, line 7: UPSTREAM_TELEMETRY_HTTP_PORT, UPSTREAM_TELEMETRY_TLS_PORT - 
Should it be part of the env variables and helm options?

FrimIdan: in general we are not configuring the ports via helm values so no need.

charts/apiclarity/templates/deployment.yaml
FrimIdan: it is on purpose for now, we need that the fuzzer will support it first and then we will change to to talk to the TLS endpoint

backend/pkg/rest/server.go
FrimIdan: If files are empty or does not exist if should fail during server creation 
```
			httpsServer.TLSConfig.Certificates[0], err = tls.LoadX509KeyPair(string(s.TLSCertificate), string(s.TLSCertificateKey))

```

fishkerez (approved):

charts/apiclarity/templates/deployment.yaml
akpsgit: worth adding a todo comment

charts/apiclarity/templates/post-delete-job-certs-deletion.yaml
```diff
@@ -0,0 +1,100 @@
...
+            - /bin/bash
+            - -c
+            - "cd /tmp &&
+{{- if .Values.trafficSource.global.httpsProxy }}
+            git config --global https.proxy {{ .Values.trafficSource.global.httpsProxy }} &&
+{{- end }}
+{{- if .Values.trafficSource.global.httpProxy }}
+            git config --global http.proxy {{ .Values.trafficSource.global.httpProxy }} &&
+{{- end }}
+            git clone -b {{ .Chart.AppVersion }} https://github.com/openclarity/apiclarity.git &&
```
akpsgit, line 97: how can this work until it's in master?

backend/pkg/traces/server.go
```diff
@@ -55,10 +65,19 @@ func CreateHTTPTracesServer(port int, traceHandleFunc HandleTraceFunc) (*HTTPTra
 
 	server.ConfigureFlags()
 	server.ConfigureAPI()
+	server.Port = config.Port
+
+	// We want to serve both http and https
+	// TODO: need to use istio to secure the http port when the wasm is sending traces
```
akpsgit, line 71: maybe this todo should be in the wasm filter config? how does this influence the traces server?

charts/apiclarity/templates/post-delete-job-certs-deletion.yaml
FrimIdan: change `{{ .Chart.AppVersion }}` to the branch name (`devel-global-specs-tls-support`)

backend/pkg/traces/server.go
FrimIdan: need to inject envoy sidecar on apiclarity deployment (the traces server)

akpsgit (approved):

PR #209, opened 2022-07-25
fix fuzzer linting

aduminuc:


backend/pkg/modules/internal/fuzzer/model/api.go
```diff
@@ -727,9 +730,15 @@ func (api *API) RetrieveInfoFromStore(ctx context.Context, accessor core.Backend
...
 				api.TestsList = append(api.TestsList, NewTest())
-				api.AddNewStatusReport(data)
+				err = api.AddNewStatusReport(data)
+				if err != nil {
+					logging.Errorf("unable to add new status report: %v", err)
+				}
+			}
+			err = api.AddNewStatusReport(data)
+			if err != nil {
+				logging.Errorf("unable to add new status report: %v", err)
```
FrimIdan, line 740: is it safe to continue here? no need a `break` of `continue`?

aduminuc:
Already fixed in Axel's branch

PR #210, opened 2022-07-25
Use chart to deploy fuzzer and stop

axelcisco:
feat: support configmap to deploy fuzzer pod


backend/pkg/modules/internal/fuzzer/fuzzer.go
```diff
@@ -132,6 +132,8 @@ func (p *pluginFuzzer) sendAPIFindingsNotification(ctx context.Context, apiID ui
 		return fmt.Errorf("failed to create 'APIFindings' notification, err=(%v)", err)
 	}
 
+	b, err := json.Marshal(notification)
+	logging.Logf("[Fuzzer] sendAPIFindingsNotification(%v): notification=%v", apiID, string(b))
```
FrimIdan, line 136: let's do it only if log.level is debug - not sure we want to marshal all the time

charts/apiclarity/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v2
 name: apiclarity
-appVersion: latest
+appVersion: 1.0
```
FrimIdan, line 3: revert

charts/apiclarity/Chart.yaml
```diff
@@ -1,8 +1,8 @@
...
 description: Charts for APIClarity deployment. Reconstruct OpenAPI Specifications from real-time workload traffic seamlessly.
-version: latest
+version: 1.0
```
FrimIdan, line 5: revert

charts/apiclarity/templates/deployment.yaml
```diff
@@ -76,11 +76,13 @@ spec:
               value: {{ .Values.apiclarity.notificationURLPrefix }}
             {{- end }}
             - name: FUZZER_DEPLOYMENT_TYPE
-              value: kubernetes
+              value: {{ .Values.apiclarity.env.plugins.FUZZER_DEPLOYMENT_TYPE }}
             - name: FUZZER_IMAGE_NAME
               value: {{ .Values.apiclarity.modules.fuzzer.docker.image }}
             - name: FUZZER_PLATFORM_HOST_FROM_FUZZER
               value: http://{{ include "apiclarity.name" . }}.{{ .Release.Namespace }}:8080/api
```
FrimIdan, line 83: `FUZZER_IMAGE_NAME` and `FUZZER_PLATFORM_HOST_FROM_FUZZER` will now be configured on the config map template, right? so no need it here?

charts/apiclarity/templates/deployment.yaml
```diff
@@ -76,11 +76,13 @@ spec:
               value: {{ .Values.apiclarity.notificationURLPrefix }}
             {{- end }}
             - name: FUZZER_DEPLOYMENT_TYPE
-              value: kubernetes
+              value: {{ .Values.apiclarity.env.plugins.FUZZER_DEPLOYMENT_TYPE }}
```
FrimIdan, line 79: is it not already being placed during 
```
          {{- range $key, $val := .Values.apiclarity.env.plugins }}
            - name: {{ $key }}
              value: {{ $val | quote }}
          {{- end }}
```
?

charts/apiclarity/templates/fuzzer-template-configmap.yaml
```diff
@@ -0,0 +1,59 @@
...
+            capabilities:
+              drop:
+              - all
+            runAsNonRoot: true
+            runAsGroup: 1001
+            runAsUser: 1001
+            privileged: false
+            allowPrivilegeEscalation: false
+          containers:
+          - name: vulnerability-scanner
```
FrimIdan, line 38: please change `vulnerability-scanner`

charts/apiclarity/templates/fuzzer-template-configmap.yaml
```diff
@@ -0,0 +1,59 @@
...
+            image: '{{ index .Values "apiclarity" "modules" "fuzzer" "docker" "image" }}'
```
FrimIdan, line 39: you can use `{{ .Values.apiclarity.modules.fuzzer.docker.image }}`

charts/apiclarity/templates/fuzzer-template-configmap.yaml
```diff
@@ -0,0 +1,59 @@
...
+            imagePullPolicy: Always
+            volumemounts:
+              - mountpath: /tmp
+                name: tmp-volume
+            env:
+            - name: PLATFORM_TYPE
+              value: "API_CLARITY"
+            - name: PLATFORM_HOST
+              value: "http://apiclarity-apiclarity:8080/api"
```
FrimIdan, line 48: Please use helm values (http://{{ include "apiclarity.name" . }}.{{ .Release.Namespace }}:8080/api).
Do we need `PLATFORM_HOST`?

charts/apiclarity/templates/fuzzer-template-configmap.yaml
```diff
@@ -0,0 +1,59 @@
...
+            - name: FUZZER
+              value: "scn-fuzzer,restler,crud"
+            - name: REQUEST_SCOPE
+              value: "global/internalservices/portshift_request"
```
FrimIdan, line 52: `portshift_request`?!

charts/apiclarity/templates/fuzzer-template-configmap.yaml
```diff
@@ -0,0 +1,59 @@
...
+            emptyDir: {}
+          securityContext:
+            capabilities:
+              drop:
+              - all
+            runAsNonRoot: true
+            runAsGroup: 1001
+            runAsUser: 1001
+            privileged: false
+            allowPrivilegeEscalation: false
```
FrimIdan, line 36: please also add `readOnlyRootFilesystem: true`

charts/apiclarity/templates/fuzzer-template-configmap.yaml
```diff
@@ -0,0 +1,59 @@
...
+              value: "scn-fuzzer,restler,crud"
+            - name: REQUEST_SCOPE
+              value: "global/internalservices/portshift_request"
+            - name: RESTLER_ROOT_PATH
+              value: "/tmp"
+            - name: RESTLER_TOKEN_INJECTOR_PATH
+              value: "/app/"
+            - name: DEBUG
+              value: true
+            resources:
```
FrimIdan, line 59: please add configurable resources like in https://github.com/openclarity/kubeclarity/blob/main/charts/kubeclarity/templates/scanner-template-configmap.yaml#L162

charts/apiclarity/templates/deployment.yaml
FrimIdan, line 79: Also, you need to create `FUZZER_DEPLOYMENT_TYPE` in the values.yaml file https://github.com/openclarity/apiclarity/blob/devel-global-specs/charts/apiclarity/values.yaml#L53

charts/apiclarity/templates/fuzzer-template-configmap.yaml
```diff
@@ -0,0 +1,59 @@
...
+            volumemounts:
+              - mountpath: /tmp
+                name: tmp-volume
+            env:
+            - name: PLATFORM_TYPE
+              value: "API_CLARITY"
+            - name: PLATFORM_HOST
+              value: "http://apiclarity-apiclarity:8080/api"
+            - name: FUZZER
+              value: "scn-fuzzer,restler,crud"
```
FrimIdan, line 50: maybe we want to make it also configurable via helm values?

charts/apiclarity/templates/fuzzer-template-configmap.yaml
```diff
@@ -0,0 +1,59 @@
...
+            - name: REQUEST_SCOPE
+              value: "global/internalservices/portshift_request"
+            - name: RESTLER_ROOT_PATH
+              value: "/tmp"
+            - name: RESTLER_TOKEN_INJECTOR_PATH
+              value: "/app/"
+            - name: DEBUG
+              value: true
```
FrimIdan, line 58: maybe we want to make it also configurable via helm values?

axelcisco:
Hi @FrimIdan. 
You are too fast :-), I hadn't added reviewer because PR was not yet ready to be reviewed. I created it to test the build. (I never tested the "draft" option when create the PR, I will do it next time)

backend/pkg/modules/internal/fuzzer/clients/configmap.go
```diff
@@ -0,0 +1,273 @@
...
+      securityContext:
+        capabilities:
+          drop:
+          - all
+        runAsNonRoot: true
+        runAsGroup: 1001
+        runAsUser: 1001
+        privileged: false
+        allowPrivilegeEscalation: false
+      readOnlyRootFilesystem: false
```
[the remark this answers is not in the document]
axelcisco: ok

backend/pkg/modules/internal/fuzzer/clients/configmap.go
```diff
@@ -0,0 +1,273 @@
...
+        cpu: "50m"
+      limits:
+        memory: "1000Mi"
+        cpu: "1000m"
+      volumes:
+        - name: tmp-volume
+          emptyDir: {}
+`)
+
+const ()
```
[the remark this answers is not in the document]
axelcisco: already done in last version

backend/pkg/modules/internal/fuzzer/clients/configmap.go
```diff
@@ -0,0 +1,273 @@
...
+
+type ConfigMapClient struct {
+	hClient            kubernetes.Interface
```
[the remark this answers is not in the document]
axelcisco: "h" is for "handle", old way to name a pointer to an object in c/c++. I will rename it.

backend/pkg/modules/internal/fuzzer/clients/configmap.go
```diff
@@ -0,0 +1,273 @@
...
+
+	// Retrieve the env var slice that will configure our pod
+	envVars := l.getEnvs(apiID, endpoint, securityItem, timeBudget)
+	logging.Logf("[Fuzzer][ConfigMapClient] envVars=%v", envVars)
+
+	// Create job struct
+	fuzzerJob, err := l.createFuzzerJob(envVars)
+	if err != nil {
+		logging.Logf("[Fuzzer][ConfigMapClient] Failed to create fuzzer job struct: %v", err)
+		return fmt.Errorf("failed to get create job struct")
```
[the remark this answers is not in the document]
axelcisco: ok

backend/pkg/modules/internal/fuzzer/clients/configmap.go
```diff
@@ -0,0 +1,273 @@
...
+	}
+
+	// Create job
+	if _, err := l.Create(fuzzerJob); err != nil {
+		logging.Logf("[Fuzzer][ConfigMapClient] Failed to create fuzzer job: %v", err)
+		return fmt.Errorf("failed to get create job")
```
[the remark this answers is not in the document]
axelcisco: ok

backend/pkg/modules/internal/fuzzer/clients/configmap.go
```diff
@@ -0,0 +1,273 @@
...
+
+func (l *ConfigMapClient) createFuzzerJob(dynEnvVars []v1.EnvVar) (*batchv1.Job, error) {
+	var job batchv1.Job
+	var fuzzerTemplate []byte
+
+	if l.configMapName == "" {
+		// Use default fuzzer job template from config map.
+		fuzzerTemplate = fuzzerJobTemplate
+	} else {
+		// Get fuzzer job template from config map.
```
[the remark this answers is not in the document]
axelcisco: I know that. 
But even if it is not used very often, nothing preventing someone to change the configmap without re-deploying everything to update future fuzzer launch.

backend/pkg/modules/internal/fuzzer/clients/configmap.go
```diff
@@ -0,0 +1,273 @@
...
+	}
+	curEnvVars := job.Spec.Template.Spec.Containers[0].Env
+	job.Spec.Template.Spec.Containers[0].Env = append(curEnvVars, dynEnvVars...)
+
+	logging.Logf("[Fuzzer][ConfigMapClient] Using job:\n%+v", job)
+
+	// Check for job name
+	if job.GetName() == "" {
+		// Manually set one
+		job.ObjectMeta.Name = jobNamePrefix + uuid.NewV4().String()
```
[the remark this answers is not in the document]
axelcisco: yes, but we are not on same condition that kubeclarity. Here jobNamePrefix is an application constants defined in ./constants.go, not a name that come from an external source (user or tag/repo). No surprise, here. We can expect that the prefix follow the rules, and if not it is something to check at build time, not at runtime

backend/pkg/modules/internal/fuzzer/clients/interface.go
```diff
@@ -54,6 +54,14 @@ func NewClient(moduleConfig *config.Config, accessor core.BackendAccessor) (Clie
 		}
 		logging.Logf("[Fuzzer] Docker Fake creation, ok")
 		return client, nil
+	} else if moduleConfig.GetDeploymentType() == config.DeploymentTypeConfigMap {
```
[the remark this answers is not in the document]
axelcisco: No. configmap and kubernetes are separated way to deploy, here.

charts/apiclarity/Chart.yaml
axelcisco: done. was commit by mistake.

backend/pkg/modules/internal/fuzzer/fuzzer.go
axelcisco: it's only for integration purpose, the trace will disappear after. But I know that apifindingsnotification already works, then I will remove it.

backend/pkg/modules/internal/fuzzer/clients/interface.go
```diff
@@ -54,6 +54,14 @@ func NewClient(moduleConfig *config.Config, accessor core.BackendAccessor) (Clie
...
+		client, err := NewConfigMapClient(moduleConfig, accessor)
+		if err != nil {
+			logging.Logf("[Fuzzer] Error, can't create Helm client, err=(%v)", err)
```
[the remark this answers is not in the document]
axelcisco: typo already fixed on a previous commit

charts/apiclarity/Chart.yaml
axelcisco: done. was commit by mistake.

backend/pkg/modules/internal/fuzzer/clients/configmap.go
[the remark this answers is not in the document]
FrimIdan: If we want to support dynamic configuration update we should watch that config map and get update events to update it during runtime. For now I think we can agree that if someone will change the config a restart for the deployment will be needed. it better then fetching it every job creation - DYT?

backend/pkg/modules/internal/fuzzer/clients/configmap.go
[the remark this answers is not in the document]
FrimIdan: understood

backend/pkg/modules/internal/fuzzer/clients/interface.go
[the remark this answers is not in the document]
FrimIdan: but why we need the hard coded "kubernetes" type?

charts/apiclarity/templates/deployment.yaml
axelcisco: No need in case of configmap use, right. FUZZER_IMAGE_NAME can be removed from this file (must be used only when starting ApiClarity in command line) FUZZER_PLATFORM_HOST_FROM_FUZZER can be moved to configmap

charts/apiclarity/templates/deployment.yaml
axelcisco: you're right

charts/apiclarity/templates/fuzzer-template-configmap.yaml
axelcisco: oups...

charts/apiclarity/templates/fuzzer-template-configmap.yaml
axelcisco: ok

charts/apiclarity/templates/fuzzer-template-configmap.yaml
axelcisco: Needed for the Fuzzer application. Mandatory for now.
The remove of all of these (old) SCN stuff will be done in MAN-9376

charts/apiclarity/templates/fuzzer-template-configmap.yaml
axelcisco: ok, done

backend/pkg/modules/internal/fuzzer/clients/configmap.go
```diff
@@ -0,0 +1,273 @@
+// Copyright © 2022 Cisco Systems, Inc. and its affiliates.
```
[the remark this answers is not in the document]
axelcisco: No, the two ways to start fuzzer, either by configmap either by creating the full job with golang k8s client exists, depending of the need

charts/apiclarity/templates/deployment.yaml
axelcisco: already done. was needed by CI / Lint Helm Chart

charts/apiclarity/templates/fuzzer-template-configmap.yaml
axelcisco: ok

backend/pkg/modules/internal/fuzzer/clients/configmap.go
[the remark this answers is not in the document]
FrimIdan: not sure why we still need the hardcoded k8s way (the job info is hardcoded in the code)
both are using go k8s client

charts/apiclarity/templates/fuzzer-template-configmap.yaml
FrimIdan: `PLATFORM_HOST` was not removed

backend/pkg/modules/internal/fuzzer/clients/configmap.go
```diff
@@ -0,0 +1,299 @@
...
+        - name: tmp-volume
+          emptyDir: {}
+`)
+
+type ConfigMapClient struct {
+	hClient            kubernetes.Interface
+	configMapName      string
+	configMapNamespace string
+	currentJob         *batchv1.Job
+	imagePullSecrets   map[int64]*tools.ImagePullSecret
```
FrimIdan, line 104: why do you call it `imagePullSecrets`? is it not `authSecrets`?

backend/pkg/modules/internal/fuzzer/tools/secret.go
```diff
@@ -0,0 +1,88 @@
...
+func (s *ImagePullSecret) Save(_ context.Context, client kubernetes.Interface) error {
+	secretDataMap := make(map[string][]byte)
+	secretDataMap[s.key] = []byte(s.body)
+
+	newSecret := corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: s.name,
+		},
+		Data: secretDataMap,
+		Type: SecretType,
```
FrimIdan, line 61: `SecretTypeOpaque is the default` no need to set it

backend/pkg/modules/internal/fuzzer/tools/secret.go
```diff
@@ -0,0 +1,88 @@
...
+
+	uuid "github.com/satori/go.uuid"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	ImagePullSecretNamePrefix = "imagepullsecret"
+	SecretType                = "Opaque"
```
FrimIdan, line 29: please remove, it is not needed.

backend/pkg/modules/internal/fuzzer/tools/secret.go
```diff
@@ -0,0 +1,88 @@
...
+	SecretKey                 = "FuzzerAuthData" // nolint:gosec
```
FrimIdan, line 30: `SecretKey` and `ImagePullSecretNamePrefix` - do you need the const to be exportable?

backend/pkg/modules/internal/fuzzer/tools/secret.go
```diff
@@ -0,0 +1,88 @@
...
+	if err != nil {
+		return err //nolint:wrapcheck // really want to return the error only
+	}
+
+	return nil
+}
+
+func (s *ImagePullSecret) Delete(_ context.Context, client kubernetes.Interface) error {
+	err := client.CoreV1().Secrets(s.namespace).Delete(context.TODO(), s.name, metav1.DeleteOptions{})
+	if err != nil {
```
FrimIdan, line 74: please add `&& !errors.IsNotFound(err)` to avoid returning a not found error

charts/apiclarity/templates/clusterrole.yaml
```diff
@@ -7,8 +7,14 @@ metadata:
...
 - apiGroups: [""]
-  resources: ["nodes", "services", "pods"]
+  resources: ["nodes", "services", "pods", "configmaps"]
+  verbs: ["get", "list", "watch", "create"]
+- apiGroups: [""]
+  resources: ["configmaps"]
   verbs: ["get", "list", "watch"]
+- apiGroups: [""]
+  resources: ["secrets"]
+  verbs: ["create", "delete"]
```
FrimIdan, line 17: better to put it in a role and not clusterrole since we don't need permissions to delete/create secrets in all namespaces, right?

There is already `charts/apiclarity/templates/role.yaml` and `charts/apiclarity/templates/rolebinding.yaml` just need to update the comment and remove the `{{- if .Values.global.traceSampling.enable -}}` & `{{- end -}}` from both files

charts/apiclarity/templates/clusterrole.yaml
```diff
@@ -7,8 +7,14 @@ metadata:
     {{ include "apiclarity.labels" . }}
 rules:
 - apiGroups: [""]
-  resources: ["nodes", "services", "pods"]
+  resources: ["nodes", "services", "pods", "configmaps"]
+  verbs: ["get", "list", "watch", "create"]
+- apiGroups: [""]
+  resources: ["configmaps"]
   verbs: ["get", "list", "watch"]
```
FrimIdan, line 14: why `configmaps` duplication and why we need `create` verb?

backend/pkg/modules/internal/fuzzer/clients/configmap.go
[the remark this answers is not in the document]
axelcisco: ok

backend/pkg/modules/internal/fuzzer/clients/interface.go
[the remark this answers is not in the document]
axelcisco: Current "kubernetes" is fully functional and can be used in parallel of "configmap" one, then in my view no need to remove it, some guys can prefer this way to deploy Fuzzer.
But you and Alexei are the owner of this project, then as it seems disturb you, I will remove it.

backend/pkg/modules/internal/fuzzer/config/config.go
```diff
@@ -71,6 +73,7 @@ const (
 	FuzzerTestReportTimeoutEnvVar    = "FUZZER_TESTREPORT_TIMEOUT"
 	FuzzerJobNamespaceEnvVar         = "FUZZER_JOB_NAMESPACE"
```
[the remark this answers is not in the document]
axelcisco: I don't know... Seems to be used :
	// FuzzerJobNamespaceEnvVar takes priority over PODNamespaceEnvVar
	jobNamespace := viper.GetString(FuzzerJobNamespaceEnvVar)
	if jobNamespace == "" {
		jobNamespace = viper.GetString(PODNamespaceEnvVar)
	}

charts/apiclarity/templates/fuzzer-template-configmap.yaml
axelcisco: done

backend/pkg/modules/internal/fuzzer/config/config.go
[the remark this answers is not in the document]
FrimIdan: it is used in `NewKubernetesClient` so if you are keeping the  `Kubernetes` deployment type you should keep also the env var, if not, all related code can be removed

charts/apiclarity/templates/fuzzer-template-configmap.yaml
axelcisco: done

backend/pkg/modules/internal/fuzzer/clients/configmap.go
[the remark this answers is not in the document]
axelcisco: see comment above.
I will remove it since it's confusing.

charts/apiclarity/templates/fuzzer-template-configmap.yaml
axelcisco: done

backend/pkg/modules/internal/fuzzer/tools/secret.go
axelcisco: I think it is related to the remark above?

backend/pkg/modules/internal/fuzzer/tools/secret.go
axelcisco: No.
I will set it no exportable.

backend/pkg/modules/internal/fuzzer/tools/secret.go
axelcisco: hmmm, yes, but "not found" error is an error. It must not occurred, otherwise it means we have an issue somewhere with the processing create/delete Secret, and this must be investigated
Are you sure you don't want to see the error concerning the Secret not found?

charts/apiclarity/templates/clusterrole.yaml
axelcisco: My bad...
I wanted to separate "configmaps" and "secret" from the "global" one as we don't need all the permissions, but I forgot to remove "configmaps" and "create" from the global line.

charts/apiclarity/templates/clusterrole.yaml
axelcisco: I have already tried to remove my stuff from ```clusterrole.yaml``` to use ```role.yaml``` by it was not a success. Note that I hadn't used ```rolebinding.yaml```.
I will try again

backend/pkg/modules/internal/fuzzer/tools/secret.go
axelcisco: what to say, here?
Perhaps it is the default , but I'm not aware. I tested various type of secret here to finish by "Opaque". In a personal mind I'm against the fashion to always remove all initializing when it is the default value. For standard type (int, string, etc...) perhaps it is relevant, for a var that have a meaning it's not so obvious and needs to go to the documentation just to read the "default" value.
If you insist I can remove this default value, WDYT?

charts/apiclarity/templates/clusterrole.yaml
axelcisco: Ok, with the update in ```rolebinding.yaml``` it works. I revert clusterrole.yaml to the original state.

backend/pkg/modules/internal/fuzzer/tools/secret.go
FrimIdan: i don't mind you to set it explicitly but IMO better to use the const defined in k8s api `v1.SecretTypeOpaque`

charts/apiclarity/templates/fuzzer-template-configmap.yaml
FrimIdan: forgot?

charts/apiclarity/templates/fuzzer-template-configmap.yaml
```diff
@@ -0,0 +1,58 @@
...
+            readOnlyRootFilesystem: true
+          containers:
+          - name: fuzzer
+            image: "{{ .Values.apiclarity.modules.fuzzer.docker.image }}"
+            imagePullPolicy: Always
+            volumeMounts:
+              - mountPath: /tmp
+                name: tmp-volume
+            env:
+            - name: PLATFORM_TYPE
```
FrimIdan, line 46: what about `FUZZER_PLATFORM_HOST_FROM_FUZZER`?

backend/pkg/modules/internal/fuzzer/clients/configmap.go
```diff
@@ -0,0 +1,318 @@
...
+	return ret, nil
+}
+
+//nolint: ireturn,nolintlint
+func NewConfigMapClient(config *config.Config, accessor core.BackendAccessor) (Client, error) {
+	client := &ConfigMapClient{
+		k8sClient:          accessor.K8SClient(),
+		platformHost:       config.GetPlatformHost(),
+		configMapName:      config.GetJobTemplateConfigMapName(),
+		configMapNamespace: config.GetJobNamespace(),
```
FrimIdan, line 301: `config.GetJobNamespace()`?

backend/pkg/modules/internal/fuzzer/tools/secret.go
FrimIdan: my take is that if I'm trying to delete something that is not found I don't have anything else to do here, but you can keep the error if you find it useful.

backend/pkg/modules/internal/fuzzer/tools/secret.go
axelcisco: ok

charts/apiclarity/templates/fuzzer-template-configmap.yaml
axelcisco: ... My bad, I thought you talked about the data DEBUG and so. Didn't realize that it was "resources".

backend/pkg/modules/internal/fuzzer/tools/secret.go
axelcisco: ok, no problem :-)

charts/apiclarity/templates/fuzzer-template-configmap.yaml
axelcisco: I Did what we discussed in team webex

FrimIdan (approved):

PR #216, opened 2022-07-26
Uniform hostname

fishkerez:
- trace sampling client that gateways use: ShouldTrace now takes port into account.
- kong plugin - port is now taken from the url.
- tyk plugin: fix getDestinationNamespaceFromHost function
- tapper - take host from URL and not from request.host

### DIFF
--- a/plugins/common/trace_sampling_client/trace_sampling_client.go
+++ b/plugins/common/trace_sampling_client/trace_sampling_client.go
@@ -74,13 +74,18 @@ func (t *Client) Start() {
 	}()
 }
 
-func (t *Client) ShouldTrace(host string) bool {
+func (t *Client) ShouldTrace(host, port string) bool {
 	t.lock.RLock()
 	defer t.lock.RUnlock()
 
 	if len(t.Hosts) == 0 {
 		return false
 	}
+
+	if port != "" {
+		host = host + ":" + port
+	}
+
 	if t.Hosts[allHosts] || t.Hosts[host] {
 		return true
 	}
--- a/plugins/common/trace_sampling_client/trace_sampling_client.go
+++ b/plugins/common/trace_sampling_client/trace_sampling_client.go
@@ -22,9 +22,9 @@ import (
 
 	log "github.com/sirupsen/logrus"
 
+	"github.com/openclarity/apiclarity/plugins/common"
 	"github.com/openclarity/trace-sampling-manager/api/client/client"
 	"github.com/openclarity/trace-sampling-manager/api/client/client/operations"
-	"github.com/openclarity/apiclarity/plugins/common"
 )
 
 type Client struct {
--- a/plugins/common/utils.go
+++ b/plugins/common/utils.go
@@ -32,6 +32,7 @@ import (
 	uuid "github.com/satori/go.uuid"
 
 	tracesamplingclient "github.com/openclarity/trace-sampling-manager/api/client/client"
+
 	"github.com/openclarity/apiclarity/plugins/api/client/client"
 	"github.com/openclarity/apiclarity/plugins/api/client/models"
 )
@@ -158,3 +159,33 @@ func createClientTransportTLS(host string, tlsOptions *ClientTLSOptions) (runtim
 
 	return transport, nil
 }
+
+func GetHostAndPortFromURL(URL string) (host, port string) {
+	if !strings.Contains(URL, "://") {
+		// need to add scheme to host in order for url.Parse to parse properly
+		URL = "http://" + URL
+	}
+
+	parsedHost, err := url.Parse(URL)
+	if err != nil {
+		return URL, ""
+	}
+
+	host = parsedHost.Hostname()
+	port = parsedHost.Port()
+
+	host = strings.TrimSuffix(host, ".svc.cluster.local")
+	host = strings.TrimSuffix(host, ".svc.cluster")
+	host = strings.TrimSuffix(host, ".svc")
+
+	if port == "" {
+		if parsedHost.Scheme == "https" {
+			port = "443"
+		} else {
+			port = "80"
+		}
+	}
+
+	return
+}
+
--- a/plugins/common/utils.go
+++ b/plugins/common/utils.go
@@ -160,7 +160,8 @@ func createClientTransportTLS(host string, tlsOptions *ClientTLSOptions) (runtim
 	return transport, nil
 }
 
-func GetHostAndPortFromURL(URL string) (host, port string) {
+// If host from URL is not containing any dots, the defaultNamespace will be added to the host name
+func GetHostAndPortFromURL(URL, defaultNamespace string) (host, port string) {
 	if !strings.Contains(URL, "://") {
 		// need to add scheme to host in order for url.Parse to parse properly
 		URL = "http://" + URL
@@ -178,6 +179,11 @@ func GetHostAndPortFromURL(URL string) (host, port string) {
 	host = strings.TrimSuffix(host, ".svc.cluster")
 	host = strings.TrimSuffix(host, ".svc")
 
+	// we assume that host with no dots is an internal host, so we will "fix" it with default namespace.
+	if !strings.Contains(host, ".") && defaultNamespace != "" {
+		host = host + "." + defaultNamespace
+	}
+
 	if port == "" {
 		if parsedHost.Scheme == "https" {
 			port = "443"
@@ -188,3 +194,16 @@ func GetHostAndPortFromURL(URL string) (host, port string) {
 
 	return
 }
+
+const (
+	MinimumSeparatedHostSize = 2
+)
+
+// Will try to extract the namespace from the host name, and if not found, will use the provided default namespace.
+func GetDestinationNamespaceFromHostOrDefault(host, defaultNamespace string) string {
+	if sp := strings.Split(host, "."); len(sp) >= MinimumSeparatedHostSize {
+		return sp[1]
+	}
+	return defaultNamespace
+}
+
--- a/plugins/common/utils.go
+++ b/plugins/common/utils.go
@@ -30,6 +30,7 @@ import (
 	httptransport "github.com/go-openapi/runtime/client"
 	"github.com/go-openapi/strfmt"
 	uuid "github.com/satori/go.uuid"
+	log "github.com/sirupsen/logrus"
 
 	tracesamplingclient "github.com/openclarity/trace-sampling-manager/api/client/client"
 
@@ -169,6 +170,7 @@ func GetHostAndPortFromURL(URL, defaultNamespace string) (host, port string) {
 
 	parsedHost, err := url.Parse(URL)
 	if err != nil {
+		log.Errorf("Failed to parse URL=%v: %v", URL, err)
 		return URL, ""
 	}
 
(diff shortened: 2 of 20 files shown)


plugins/gateway/kong/plugin.go
```diff
@@ -113,11 +113,11 @@ func shouldTrace(kong *pdk.PDK) (bool, error) {
...
 		return false, fmt.Errorf("failed to get routed service: %v", err)
 	}
-	host, _ := parseHost(routedService.Host)
-	if traceSamplingClient.ShouldTrace(host) {
+	host, port, _ := parseKongHost(routedService.Host)
+	if traceSamplingClient.ShouldTrace(host, port) {
 		return true, nil
 	}
-	_ = kong.Log.Info("Ignoring host: %v", host)
+	_ = kong.Log.Info("Ignoring host: %v:%v", host, port)
```
klyr, line 120: Are we showing this log, with INFO level, for each and every ignored traces ?

plugins/gateway/kong/deploy/deploy.sh
```diff
@@ -4,13 +4,13 @@ KongGatewayDeploymentName="${KONG_GATEWAY_DEPLOYMENT_NAME:-kong}"
...
-UpstreamTelemetryHostName="${UPSTREAM_TELEMETRY_HOST_NAME:-apiclarity-apiclarity.apiclarity:9000}"
-TraceSamplingHostName="${TRACE_SAMPLING_HOST_NAME:-apiclarity-apiclarity.apiclarity:9990}"
+UpstreamTelemetryAddress="${UPSTREAM_TELEMETRY_ADDRESS:-apiclarity-apiclarity.apiclarity:9000}"
+TraceSamplingAddress="${TRACE_SAMPLING_ADDRESS:-apiclarity-apiclarity.apiclarity:9990}"
 TraceSamplingEnabled="${TRACE_SAMPLING_ENABLED:-false}"
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 
-cat "${DIR}/kongPlugin.yaml" | sed "s/{{UPSTREAM_TELEMETRY_HOST_NAME}}/$UpstreamTelemetryHostName/g" | sed "s/{{TRACE_SAMPLING_HOST_NAME}}/$TraceSamplingHostName/g" | sed "s/{{TRACE_SAMPLING_ENABLED}}/$TraceSamplingEnabled/g" | kubectl -n ${KongGatewayIngressNamespace} apply -f -
+cat "${DIR}/kongPlugin.yaml" | sed "s/{{UPSTREAM_TELEMETRY_ADDRESS}}/$UpstreamTelemetryAddress/g" | sed "s/{{TRACE_SAMPLING_ADDRESS}}/$TraceSamplingAddress/g" | sed "s/{{TRACE_SAMPLING_ENABLED}}/$TraceSamplingEnabled/g" | kubectl -n ${KongGatewayIngressNamespace} apply -f -
```
klyr, line 13: For fun ;)
```shell
sed 's/{{UPSTREAM_TELEMETRY_ADDRESS}}/$UpstreamTelemetryAddress/g; s/{{TRACE_SAMPLING_ADDRESS}}/$TraceSamplingAddress/g; s/{{TRACE_SAMPLING_ENABLED}}/$TraceSamplingEnabled/g' ${DIR}/kongPlugin.yaml | kubectl -n ${KongGatewayIngressNamespace} apply -f -
```

plugins/gateway/kong/README.md
```diff
@@ -56,7 +56,7 @@ If you just want to try it out with a demo application, and you don't have kong
        KONG_GATEWAY_DEPLOYMENT_NAMESPACE=kong \
        KONG_GATEWAY_INGRESS_NAME=catalogue \
        KONG_GATEWAY_INGRESS_NAMESPACE=sock-shop \
-       UPSTREAM_TELEMETRY_HOST_NAME=apiclarity-apiclarity.apiclarity:9000 \
+       UPSTREAM_TELEMETRY_ADDRESS=apiclarity-apiclarity.apiclarity:9000 \
```
klyr, line 59: Is documentation needed for `TRACE_SAMPLING_ENABLED` and  `TRACE_SAMPLING_ADDRESS` ?

plugins/taper/run.go
```diff
@@ -134,8 +134,7 @@ func (a *Agent) startReadOutputItems(ctx context.Context, outputItems chan *api.
 					log.Errorf("Failed to create telemetry: %v", err)
 					return
 				}
-				// TODO format host to host.namespace (or whatever format we will decide on for hosts)
-				if !a.shouldTrace(telemetry.Request.Host) {
+				if !a.shouldTrace(telemetry.Request.Host, item.ConnectionInfo.ServerPort) {
 					log.Infof("Ignoring host: %v", telemetry.Request.Host)
```
klyr, line 138: Same question, do we really want to log in Info level every traces that are ignored ?

plugins/common/trace_sampling_client/trace_sampling_client.go
```diff
@@ -74,13 +74,18 @@ func (t *Client) Start() {
...
-func (t *Client) ShouldTrace(host string) bool {
+func (t *Client) ShouldTrace(host, port string) bool {
 	t.lock.RLock()
 	defer t.lock.RUnlock()
 
 	if len(t.Hosts) == 0 {
 		return false
 	}
+
+	if port != "" {
```
klyr, line 85: Apparently, it's also possible to have `*` for the port.
Need to align with @aduminuc about that. (see https://github.com/openclarity/apiclarity/blob/devel-global-specs/backend/pkg/modules/internal/core/core.go#L116)

klyr (changes requested):

klyr (commented):
That looks good to me.

plugins/gateway/kong/plugin.go
```diff
@@ -113,11 +113,11 @@ func shouldTrace(kong *pdk.PDK) (bool, error) {
 	if err != nil {
 		return false, fmt.Errorf("failed to get routed service: %v", err)
 	}
-	host, _ := parseHost(routedService.Host)
-	if traceSamplingClient.ShouldTrace(host) {
+	host, port, _ := parseKongHost(routedService.Host)
```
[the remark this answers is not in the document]
fishkerez: Not sure. 
We use kong.Router.GetService().Host(), which the docs state:
`// kong.Router.GetService() returns the current service entity.
// The request will be targetted to this upstream service.`

On the other hand there is no direct access to the URL, just to parts of the URL via:
 kong.Request.GetForwardedPort() and kong.Request.GetForwardedHost()
 
 I will check these to see if it is working and maybe use them, as it seems more correct (again, not sure)

plugins/gateway/kong/plugin.go
fishkerez: Yes

plugins/gateway/kong/deploy/deploy.sh
fishkerez: Looks really fun I admit, but what is not broken, do not fix :)

plugins/gateway/kong/README.md
fishkerez: Probably yes. Not part of this task though

plugins/taper/run.go
fishkerez: I think that if you not running in Info level (which you dont by default), It is a NOOP. If you think it is a crucial thing, we can think of not printing here, or change it to TRACE

plugins/common/trace_sampling_client/trace_sampling_client.go
fishkerez: Not sure, Is it supported also in Trace Sampling Manager?

klyr: After discussion with @aduminuc, it may be supported in the future. And is already supported in APIClarity.
So, no need to take that into accound right now.

All good.

klyr (approved):

plugins/gateway/kong/plugin.go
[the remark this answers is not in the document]
fishkerez: I think it is best to take the host and port from the routing information after all, since these are for sure the target service taken from the Ingress resource

plugins/common/utils.go
```diff
@@ -158,3 +159,51 @@ func createClientTransportTLS(host string, tlsOptions *ClientTLSOptions) (runtim
...
+			port = "80"
+		}
+	}
+
+	return
+}
+
+const (
+	MinimumSeparatedHostSize = 2
+)
```
FrimIdan, line 200: move with all other const

plugins/taper/run_test.go
```diff
@@ -27,6 +27,7 @@ func TestAgent_shouldTrace(t *testing.T) {
 	}
 	type args struct {
 		host string
+		port string
```
akpsgit, line 30: worth adding the port also in test args

plugins/taper/run.go
```diff
@@ -211,17 +212,20 @@ func (a *Agent) createTelemetry(item *api.OutputChannelItem) (*models.Telemetry,
 
 	pathAndQuery := common.GetPathWithQuery(request.URL)
 
+	host, _ := common.GetHostAndPortFromURL(request.URL.String(), a.config.TapperNamespace)
+	destinationNamespace := common.GetDestinationNamespaceFromHostOrDefault(host, a.config.TapperNamespace)
+
```
FrimIdan, line 217: `a.config.TapperNamespace` is not the great default namespace to use I think.
We need to use the client namespace so if the host does not hold `.namespace` we can assume that the service is in the same namespace of the **client** and not the tapper.

fishkerez: You right

plugins/common/utils_test.go
```diff
@@ -153,3 +153,93 @@ func TestCreateHeaders(t *testing.T) {
...
+		{
+			name: "with port",
+			args: args{
+				url: "http://catalogue.sock-shop:8080",
+			},
+			wantHost: "catalogue.sock-shop",
+			wantPort: "8080",
+		},
+		{
+			name: "with port - no namespace - use default",
```
FrimIdan, line 185: need to add a case for external host `www.example.com` with `defaultNamespace` set

plugins/taper/run.go
```diff
@@ -48,6 +48,7 @@ type Agent struct {
 	podMonitor          *monitor.PodMonitor
 	apiClient           *client.APIClarityPluginsTelemetriesAPI
 	traceSamplingClient *trace_sampling_client.Client
+	config              *config.Config
```
FrimIdan, line 51: why it is needed?

FrimIdan (approved):

plugins/gateway/kong/plugin.go
```diff
@@ -237,18 +234,25 @@ func getRequestTimeFromContext(kong *pdk.PDK) (int64, error) {
 	return int64(requestTime), nil
 }
 
-// KongHost: <svc-name>.<namespace>.8000.svc
-// convert to name.namespace.
-func parseHost(kongHost string) (host, namespace string) {
+// KongHost format: <svc-name>.<namespace>.8000.svc.
+func parseKongHost(kongHost string) (host, port, namespace string) {
```
akpsgit, line 238: worth adding a unit test

PR #227, opened 2022-07-28
UI Findings Component

michrya2:
- adds a new findings component to handle updated findings response
format from backend.


klyr (approved):
I tested it.
This looks good to me.
I can't comment the code, but the UI seems working well.

PR #238, opened 2022-08-25
TA: Adds additional_info for same password

klyr:


backend/pkg/modules/internal/traceanalyzer/traceanalyzer.go
```diff
@@ -409,7 +409,8 @@ func fromCoreAPIAnnotation(coreAnn *core.Annotation) (anns []utils.TraceAnalyzer
 			anns = append(anns, a)
 		}
 	}
-	return
+
+	return anns, nil
```
aduminuc, line 413: Was this a bug?

backend/pkg/modules/internal/traceanalyzer/weakbasicauth/findings.go
```diff
@@ -241,6 +241,12 @@ func (a APIAnnotationSamePassword) Redacted() utils.TraceAnalyzerAPIAnnotation {
 }
 
 func (a *APIAnnotationSamePassword) ToAPIFinding() oapicommon.APIFinding {
+	var additionalInfo *map[string]interface{}
+	if len(a.APIs) > 0 {
+		additionalInfo = &map[string]interface{}{
+			"apis": a.APIs,
```
aduminuc, line 247: Are you keeping the list of affected APIs in the description and in the additional Info?

aduminuc (approved):

backend/pkg/modules/internal/traceanalyzer/weakbasicauth/findings.go
klyr: Yes.

PR #239, opened 2022-08-26
TA: Adds additional_info for sensitive JWT claims

klyr:


backend/pkg/modules/internal/traceanalyzer/weakjwt/findings.go
```diff
@@ -573,6 +586,10 @@ func (a APIAnnotationSensitiveContent) Redacted() utils.TraceAnalyzerAPIAnnotati
 }
 
 func (a *APIAnnotationSensitiveContent) ToAPIFinding() oapicommon.APIFinding {
+	additionalInfo := &map[string]interface{}{
+		"sensitive_jwt_claims":  common_utils.MapToSlice(a.SensitiveWordsInClaims),
+		"sensitive_jwt_headers": common_utils.MapToSlice(a.SensitiveWordsInHeaders),
```
aduminuc, line 591: Are you disclosing any sensitive information with this additional_info?

aduminuc (approved):

backend/pkg/modules/internal/traceanalyzer/weakjwt/findings.go
klyr: No, I'm only looking at the key of the claims, not its value.
Some, if I have a JWT claim like
```json
{
  "password": "veryunsecure"
}
```

I only report "password" and not the password itself. This finding is here to say that if you name a claim "password" (or any other sensitive word), you are certainly doing something wrong.

PR #241, opened 2022-08-29
Devel global specs path id in core

fishkerez:


backend/pkg/backend/backend.go
```diff
@@ -318,6 +318,21 @@ func (b *Backend) handleHTTPTrace(ctx context.Context, trace *pluginsmodels.Tele
 		}
 	}
 
+	var providedPathID string
+	var reconstructedPathID string
+	if b.speculator.HasProvidedSpec(specKey) {
+		providedPathID, err = b.speculator.GetPathID(specKey, telemetry.Request.Path, _spec.SpecSourceProvided)
```
FrimIdan, line 324: need to use `path` from `_spec.GetPathAndQuery` and not `telemetry.Request.Path` when calling to b.speculator.GetPathID

FrimIdan (approved):

PR #247, opened 2022-09-09
Fill provided_spec_location field for Fuzzer findings

axelcisco:
Fill the provided_spec_location field on Fuzzer findings


backend/pkg/modules/internal/fuzzer/model/api.go
```diff
@@ -350,6 +352,11 @@ func updateRequestCounter(shortReport *restapi.ShortTestReport, path string, ver
 	return fmt.Errorf("can't find operation(%v %v) in spec to update requests counter", verb, path)
 }
 
+func getLocation(path string, method string) *string {
```
klyr, line 355: Why do you need to return a pointer to string ?

PR #255, opened 2022-09-20
Otel collector plugin

jnapper7:
This PR adds a plugin for the [OpenTelemetry Collector](https://github.com/open-telemetry/opentelemetry-collector). Specifically, it adds an exporter for the collector that exports OTel traces that are server/client to APIClarity in the Telemetry format. The plugin dir includes build of a docker image of the otel collector with the addition of the APIClarity exporter plugin.


jeff-trianglecm:
1.	How will release happen? what is being built and what the user need to build? What the Dockerfile is for?

    The Dockerfile builds the OpenTelemetry collector with the repo's APIClarity exporter included. The OTel collector plugins are golang packages. The Dockerfile pulls a recent OTel collector config and adds in the APIClarity exporter, rebuilding everything to additionally include the exporter (and everything else already included in the default OTel collector).
    
    Eventually the apiclarityexporter package should be available on (the public Go package list)[https://pkg.go.dev/] because I would like to contribute the exporter to the main OTel collector repo. When it is added there, the APIClarity exporter would be generally available and you would not need to build your own version; you could simply configure the default version.
    
    The container and the apiclarityexporter package are the only build artefacts.
  
2.	Should this be part of Helm installation/configuration?

    No, I don't believe so. The plugin would only be configured in the OTel collector's configuration. An example of that config is here: [https://github.com/openclarity/apiclarity/blob/otel-collector-plugin/plugins/otel-collector/configs/otel-collector-config.yaml](https://github.com/openclarity/apiclarity/blob/otel-collector-plugin/plugins/otel-collector/configs/otel-collector-config.yaml).

3.	Should trace sampling manager be integrated into the otel exporter?

    Yes, maybe? Sampling is already handled by the tracing infrastructure, both at the SDK level where tracing is generated and within the OTel collector itself by selecting different traces. It might be confusing to the user if the exporter also subsampled, but the config for that sampling was outside of the OTel configuration. Then again, maybe not?

4.	what are the "replaces" commits in the readme?

    Does this refer to the instructions on building the OTel collector at the bottom? These are used to build the OTel collector using the APIClarity exporter plugin package. Since there is no public version of the package yet, it has to be built locally using the local package. The replacement for the APIClarity api package is due to the crazy build packages of APIClarity itself.

5.	Should be added to the main readme of APIClarity?

    I'm happy to add it as one of the example source integrations, which is where it would be relevant.

FrimIdan:
> 1. The container and the apiclarityexporter package are the only build artefacts.

So, do you want/need to the container image during the [release flow](https://github.com/openclarity/apiclarity/blob/master/.github/workflows/release.yaml) as we are doing to all other containers? 

> Eventually the apiclarityexporter package should be available on (the public Go package list)[https://pkg.go.dev/]

How it will become public? Once it will be pushed to master and tagged we can remove all the replaces in the readme?

jeff-trianglecm:
> > 1. The container and the apiclarityexporter package are the only build artefacts.
> 
> So, do you want/need to the container image during the [release flow](https://github.com/openclarity/apiclarity/blob/master/.github/workflows/release.yaml) as we are doing to all other containers?

I think releasing the docker container makes sense along with the normal release flow so it should be added to the release action. Do you want me to do that?

> > Eventually the apiclarityexporter package should be available on (the public Go package list)[https://pkg.go.dev/]
> 
> How it will become public? Once it will be pushed to master and tagged we can remove all the replaces in the readme?

Yes, I think so.

jnapper7:
> > 1. The container and the apiclarityexporter package are the only build artefacts.
> 
> So, do you want/need to the container image during the [release flow](https://github.com/openclarity/apiclarity/blob/master/.github/workflows/release.yaml) as we are doing to all other containers?

I think releasing the docker container makes sense along with the normal release flow so it should be added to the release action. Do you want me to do that?

> > Eventually the apiclarityexporter package should be available on (the public Go package list)[https://pkg.go.dev/]
> 
> How it will become public? Once it will be pushed to master and tagged we can remove all the replaces in the readme?

Yes, I think so.

FrimIdan:
> > > 1. The container and the apiclarityexporter package are the only build artefacts.
> > 
> > 
> > So, do you want/need to the container image during the [release flow](https://github.com/openclarity/apiclarity/blob/master/.github/workflows/release.yaml) as we are doing to all other containers?
> 
> I think releasing the docker container makes sense along with the normal release flow so it should be added to the release action. Do you want me to do that?
> 

Yes please, it should be added to `docker.yaml` and `release.yaml` workflow

> > > Eventually the apiclarityexporter package should be available on (the public Go package list)[https://pkg.go.dev/]
> > 
> > 
> > How it will become public? Once it will be pushed to master and tagged we can remove all the replaces in the readme?
> 
> Yes, I think so.

OK, understand, so once it will be pushed and tagged we can check if that works and update the docs to remove the `replace` stuff.

Makefile
```diff
@@ -106,13 +107,15 @@ lint: bin/golangci-lint ## Run linter
...
 	cd plugins/taper && ../../bin/golangci-lint run
+	$(MAKE) -C plugins/otel-collector lint
 
 .PHONY: fix
 fix: bin/golangci-lint ## Fix lint violations
 	cd backend && ../bin/golangci-lint run --fix
 	cd plugins/gateway/kong && ../../../bin/golangci-lint run --fix
 	cd plugins/gateway/tyk/v3.2.2 && ../../../../bin/golangci-lint run --fix
 	cd plugins/taper && ../../bin/golangci-lint run --fix
+	$(MAKE) -C plugins/otel-collector fix
```
FrimIdan, line 118: @jnapper7 can you please also add it under `license-check`?

jnapper7: Done!

jnapper7:
> > > > 1. The container and the apiclarityexporter package are the only build artefacts.
> > > 
> > > 
> > > So, do you want/need to the container image during the [release flow](https://github.com/openclarity/apiclarity/blob/master/.github/workflows/release.yaml) as we are doing to all other containers?
> > 
> > 
> > I think releasing the docker container makes sense along with the normal release flow so it should be added to the release action. Do you want me to do that?
> 
> Yes please, it should be added to `docker.yaml` and `release.yaml` workflow

The last commit finally does this; however, I cannot really test it myself. :(

> > > > Eventually the apiclarityexporter package should be available on (the public Go package list)[https://pkg.go.dev/]
> > > 
> > > 
> > > How it will become public? Once it will be pushed to master and tagged we can remove all the replaces in the readme?
> > 
> > 
> > Yes, I think so.
> 
> OK, understand, so once it will be pushed and tagged we can check if that works and update the docs to remove the `replace` stuff.

.github/workflows/docker.yml
```diff
@@ -87,4 +90,13 @@ jobs:
...
           build-args: |
-            VERSION=latest
\ No newline at end of file
+            VERSION=latest
+
+      - name: Build OpenTelemetry Collector with APIClarity plugin
+        uses: docker/build-push-action@v3
+        with:
+          context: plugins
+          tags: otel-apiclarityexporter:latest,otel-apiclarityexporter:$GITHUB_SHA,otel-apiclarityexporter:$OTEL_COLLECTOR_VERSION
```
FrimIdan, line 99: tags should be in the format of `ghcr.io/openclarity/otel-apiclarityexporter:<tag>`

FrimIdan, line 99: you are missing the `file` (docker file path), `cache-from` and `cache-to`. you can look at `Build passive taper`

.github/workflows/release.yaml
```diff
@@ -160,6 +161,15 @@ jobs:
...
 
+      - name: Build and Push OpenTelemetry Collector with APIClarity plugin
+        uses: docker/build-push-action@v3
+        with:
+          context: plugins
+          tags: otel-apiclarityexporter:${{ github.event.inputs.version }}
+          push: true
+          build-args: |
+            OTEL_COLLECTOR_VERSION=$OTEL_COLLECTOR_VERSION
+
```
FrimIdan, line 172: same comments as in the docker.yaml

plugins/otel-collector/Dockerfile
```diff
@@ -0,0 +1,33 @@
+FROM golang:1.18-alpine3.14 AS otelcol
+
+RUN apk add --update --no-cache gcc g++
+ARG OTEL_COLLECTOR_VERSION=0.60.0
```
FrimIdan, line 4: shouldn't `OTEL_COLLECTOR_VERSION` be empty and get it only from build-args or build-args overwrite it?

FrimIdan:
> The last commit finally does this; however, I cannot really test it myself. :(
> 
> > > > >

Best way to check it is to fork the project to your local account, then you can test any action with any condition (you can run the build action on any push to your side-branch for example)

.github/workflows/release.yaml
jnapper7: In [76ce43d](https://github.com/openclarity/apiclarity/pull/255/commits/76ce43d1be9f34c99a5d51006373dc5414768c1c)

.github/workflows/docker.yml
jnapper7: In [76ce43d](https://github.com/openclarity/apiclarity/pull/255/commits/76ce43d1be9f34c99a5d51006373dc5414768c1c)

jnapper7: In [76ce43d](https://github.com/openclarity/apiclarity/pull/255/commits/76ce43d1be9f34c99a5d51006373dc5414768c1c)

.github/workflows/docker.yml
```diff
@@ -87,4 +90,13 @@ jobs:
           cache-from: type=local,src=/tmp/.buildx-cache
           cache-to: type=local,dest=/tmp/.buildx-cache
           build-args: |
-            VERSION=latest
\ No newline at end of file
+            VERSION=latest
+
+      - name: Build OpenTelemetry Collector with APIClarity plugin
+        uses: docker/build-push-action@v3
```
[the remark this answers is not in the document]
jnapper7: In [76ce43d](https://github.com/openclarity/apiclarity/pull/255/commits/76ce43d1be9f34c99a5d51006373dc5414768c1c)

plugins/otel-collector/Dockerfile
jnapper7: I think the arg there is a default value when it isn't specified by the build command. Do you prefer no default? It's also a good method to force a failure when there is no argument in the build command to make sure you have the argument in the build command. :)

FrimIdan: I don't mind to have a default.

.github/workflows/docker.yml
FrimIdan: you are missing the org name (openclarity) in `ghcr.io/otel-apiclarityexporter:latest`

jnapper7:
> > The last commit finally does this; however, I cannot really test it myself. :(
> > > > > >
> 
> Best way to check it is to fork the project to your local account, then you can test any action with any condition (you can run the build action on any push to your side-branch for example)

Ok, I've checked the docker workflow, but the release workflow is stuck for the same reason as the above CI/build problem. :(

FrimIdan:
> > > The last commit finally does this; however, I cannot really test it myself. :(
> > > > > > >
> > 
> > 
> > Best way to check it is to fork the project to your local account, then you can test any action with any condition (you can run the build action on any push to your side-branch for example)
> 
> Ok, I've checked the docker workflow, but the release workflow is stuck for the same reason as the above CI/build problem. :(

Saw the problem, not sure why it is running lint on non relevant code, did you tried to run it locally?
Can you maybe try to remove the use of `make -C ...` and do
```
cd plugins/otel-collector/apiclarityexporter && ../../../bin/golangci-lint run
```

jnapper7:
> > > > The last commit finally does this; however, I cannot really test it myself. :(
> > > > > > > >
> > > 
> > > 
> > > Best way to check it is to fork the project to your local account, then you can test any action with any condition (you can run the build action on any push to your side-branch for example)
> > 
> > 
> > Ok, I've checked the docker workflow, but the release workflow is stuck for the same reason as the above CI/build problem. :(
> 
> Saw the problem, not sure why it is running lint on non relevant code, did you tried to run it locally? Can you maybe try to remove the use of `make -C ...` and do
> 
> ```
> cd plugins/otel-collector/apiclarityexporter && ../../../bin/golangci-lint run
> ```

Indeed. I have tried so many things. :)

At the moment it passes b/c I specifically include the broken package that doesn't pass typecheck. I did align the Makefiles so they should all be the same now.

Let me know if ignoring the one package for now is fine or not. I really have tried everything that I could think of (local run, changing the package version, etc.).

jnapper7:
> > The last commit finally does this; however, I cannot really test it myself. :(
> > > > > >
> 
> Best way to check it is to fork the project to your local account, then you can test any action with any condition (you can run the build action on any push to your side-branch for example)

I have verified this in a fork in my own repo here: [Forked Actions](https://github.com/jnapper7/apiclarity/actions)
Note the Release action only fails to publish Helm, which I specifically didn't authorize b/c I naturally didn't want to publish it. :)

FrimIdan:
> Let me know if ignoring the one package for now is fine or not

It's fine by me

FrimIdan:
> 5\. I'm happy to add it as one of the example source integrations, which is where it would be relevant.

We are missing that one and we probably can ship it

FrimIdan (approved):

PR #259, opened 2022-09-27
Adds dataframe

klyr:


backend/pkg/modules/internal/analytics_core/init.go
```diff
@@ -40,17 +44,19 @@ const (
...
+	numWorkers           int
+	proccFuncRegistered  map[TopicType][]AnalyticsModuleProccFunc
+	dataFramesRegistered map[AnalyticsModuleProccFunc]*ProcFuncDataFrames // Each registered function have a correponding Dataframes
+	topics               []TopicType
 }
 
+
 type ProcFuncDataFrames struct {
-	dataFrames map[int]*interface{}
+	dataFrames map[int]dataframe.DataFrame
```
mginzbur, line 59: The idea is that each AnalyticsModuleProccFunc may have more than one data frame. So we shall have something like map[string] map[int] dataframe.DataFrame...

Or better add a structure layer to encapsulate the internal map.

backend/pkg/modules/internal/analytics_core/init.go
```diff
@@ -151,6 +129,18 @@ func (p *AnalyticsCore) RegisterAnalyticsModuleHandler(topic TopicType, proccFun
 
 	p.proccFuncRegistered[topic] = append(p.proccFuncRegistered[topic], proccFunc)
 	p.proccFuncRegistered[topic] = orderHandlerFuncsByPriority(p.proccFuncRegistered[topic])
+
+	// Adds dataframes for this new registered function
+	p.dataFramesRegistered[proccFunc] = &ProcFuncDataFrames{dataFrames: map[int]dataframe.DataFrame{}}
```
mginzbur, line 134: The data frames shall not be automatically generated, but rather generated upon registration of data frame for ProcessingFunction

backend/pkg/modules/internal/analytics_core/init.go
```diff
@@ -46,16 +55,42 @@ type AnalyticsCore struct {
...
+	dataFrames map[DataFrameID]DataFrame
+}
+
+func NewDataFramesRegistry(numWorkers int) DataFramesRegistry {
+	return DataFramesRegistry{
+		dataFrames: map[DataFrameID]DataFrame{},
+	}
+}
+
+func (dfr DataFramesRegistry) NewDataFrame(name DataFrameID, shards int) error {
```
mginzbur, line 72: I think we need here as a parameter size limit of the data frame and optionally TTL

backend/pkg/modules/internal/analytics_core/init.go
```diff
@@ -46,16 +55,42 @@ type AnalyticsCore struct {
...
+
+func (dfr DataFramesRegistry) Get(name DataFrameID) (DataFrame, bool) {
+	df, found := dfr.dataFrames[name]
+	return df, found
 }
 
 type AnalyticsModuleProccFunc interface {
 	GetPriority() int
-	ProccFunc(topicName TopicType, dataFrames *ProcFuncDataFrames, partitionID int, message pubsub.MessageForBroker, annotations []interface{}, handler *AnalyticsCore) (newAnnotations []interface{})
+	ProccFunc(topicName TopicType, dataFramesRegistry DataFramesRegistry, partitionID int, message pubsub.MessageForBroker, annotations []interface{}, handler *AnalyticsCore) (newAnnotations []interface{})
```
mginzbur, line 93: The only problem with such a design is that it is not safe, i.e it is possible to access data frames that belong to other workers and/or other shards... Meanwhile it is not critical and I prefer just to continue with that. 

By the way. Atleast the enforcement of the not sharing could easily be done. You can add to Get function the name of the processing function and also add it to the NewDataFrame...

It is not critical at this stage, I think
..

backend/pkg/modules/internal/analytics_core/init.go
```diff
@@ -72,52 +110,39 @@ func (p *AnalyticsCore) handlerFunction(topic TopicType, partitionID int, msgCha
...
-		info:                &core.ModuleInfo{Name: "analytics_core", Description: "analytics_core"},
-		numWorkers:          1,
-		proccFuncRegistered: map[TopicType][]AnalyticsModuleProccFunc{},
-		topics:              make([]TopicType, 0, maxNumTopics),
-	}
-	/* We do not need to expose API at this point
-	handler := &httpHandler{
-		accessor: accessor,
+func (p *AnalyticsCore) getDataFramesShardsForFunc(function AnalyticsModuleProccFunc, partitionID int) map[DataFrameID]dataframe.DataFrame {
+	registeredDfs := p.proccFuncDataframeRegistered[function]
```
mginzbur, line 121: Need to handle situation where there is no entry for a function

backend/pkg/modules/internal/analytics_core/init.go
```diff
@@ -153,6 +178,15 @@ func (p *AnalyticsCore) RegisterAnalyticsModuleHandler(topic TopicType, proccFun
 	p.proccFuncRegistered[topic] = orderHandlerFuncsByPriority(p.proccFuncRegistered[topic])
 }
 
+func (p *AnalyticsCore) RegisterDataFrameForFunc(proccFunc AnalyticsModuleProccFunc, dataframe DataFrameID) {
+	registeredDFs, found := p.proccFuncDataframeRegistered[proccFunc]
```
mginzbur, line 182: Following previous comment you need to use proccFunc.GetName() as a key here

backend/pkg/modules/internal/analytics_core/init.go
mginzbur: also following previous comment need to use here function.GetName() as a key

backend/pkg/modules/internal/analytics_core/init.go
```diff
@@ -37,25 +43,57 @@ const (
...
 
-type ProcFuncDataFrames struct {
-	dataFrames map[int]*interface{}
+func (dfr DataFramesRegistry) Get(name DataFrameID) (DataFrame, bool) {
+	df, found := dfr.dataFrames[name]
+	return df, found
 }
 
 type AnalyticsModuleProccFunc interface {
 	GetPriority() int
```
mginzbur, line 95: Add here GetName() string 


To be used as a key for dataframe registration

backend/pkg/modules/internal/analytics_core/init.go
```diff
@@ -37,25 +43,57 @@ const (
...
-	numWorkers          int
-	proccFuncRegistered map[TopicType][]AnalyticsModuleProccFunc
-	topics              []TopicType
+	httpHandler                  http.Handler
+	msgBroker                    *pubsub.Handler
+	accessor                     core.BackendAccessor
+	info                         *core.ModuleInfo
+	numWorkers                   int
+	proccFuncRegistered          map[TopicType][]AnalyticsModuleProccFunc
+	proccFuncDataframeRegistered map[AnalyticsModuleProccFunc]map[DataFrameID]bool
```
mginzbur, line 59: change to map[AnalyticsModuleProccFuncName] map[DataFrameID] bool

backend/pkg/modules/internal/analytics_core/init.go
klyr: When there is not entry, the nil map is returned. In that case, the returned dfShards is empty, because we didn't enter into the for loop

mginzbur (approved):

PR #260, opened 2022-09-30
Aduminuc cross build

aduminuc:


FrimIdan:
@aduminuc Not sure I understand why it is needed, the images that is being built via `make push-docker-backend` from Mac can run on k8s AFAIK

aduminuc:
> @aduminuc Not sure I understand why it is needed, the images that is being built via `make push-docker-backend` from Mac can run on k8s AFAIK. By default `docker build` build images for `linux/amd64` platform.

@FrimIdan I just got a new M1 Mac and in my case  the docker images are built for arm and they won't run on k8s.

FrimIdan:
> @FrimIdan I just got a new M1 Mac and in my case the docker images are built for arm and they won't run on k8s.

@aduminuc `export DOCKER_DEFAULT_PLATFORM=linux/amd64` is not working?
https://docs.docker.com/engine/reference/commandline/cli/

PR #261, opened 2022-10-02
bump labstack/echo to v4.9.0

akpsgit:
fix CVE-2022-40083

### DIFF
--- a/backend/go.mod
+++ b/backend/go.mod
@@ -138,4 +138,5 @@ replace github.com/openclarity/apiclarity/api v0.0.0 => ./../api
 replace github.com/openclarity/apiclarity/plugins/api v0.0.0 => ./../plugins/api
 
 // fix CVE-2022-40083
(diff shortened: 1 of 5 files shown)


backend/go.mod
```diff
@@ -76,7 +76,7 @@ require (
 	github.com/jinzhu/now v1.1.2 // indirect
 	github.com/josharian/intern v1.0.0 // indirect
 	github.com/json-iterator/go v1.1.12 // indirect
-	github.com/labstack/echo/v4 v4.6.3 // indirect
```
FrimIdan, line 79: it is an indirect dep, if you run go mod tidy it will be reverted, no?

akpsgit: No :(

akpsgit: Noooooooo

FrimIdan (approved):

PR #270, opened 2022-10-11
New discovered apis api

klyr:


api/swagger.yaml
```diff
@@ -800,6 +800,31 @@ paths:
         default:
           $ref: '#/responses/UnknownError'
 
+  /control/newDiscoveredAPIs:
```
akpsgit, line 803: maybe externalDiscoveryAPIs?

klyr: I really meant internal, I fixed it in another commit.
So, /control/internalDiscoveryAPIs.

This new /control/ endpoint is meant to be used by Gateways, in order to notify APIClarity of new APIs.
And in general, /control/ will group all the APIs of APIClarity "control-plane".

akpsgit: cool :) will other control APIs also move under /control?
If it's meant for gateways, maybe we can add it explicitly to the path (e.g. /control/gateway/internalDiscoveryAPIs.)?
I'm less a fan of putting in the description who is the consumer of the API (maybe as example)

api/swagger.yaml
```diff
@@ -800,6 +800,31 @@ paths:
...
+    post:
+      summary: 'List of new external services discovered by API gateways'
```
aduminuc, line 805: The description should say, this method allows a client (a gateway for example) to notify APIclarity about newly discovered APIs.

In the description I would also  make sure it's clear that this can be a cumulative list or delta, which means that if the list contains already existing APIS, then they are ignored (or updated?)

backend/pkg/rest/control_controller.go
```diff
@@ -0,0 +1,65 @@
...
+		if err != nil {
+			log.Errorf("In '%s', port '%s' is invalid", h, strPort)
+			continue
+		}
+
+		apiInfo := &_database.APIInfo{
+			Type:                 models.APITypeINTERNAL,
+			Name:                 host,
+			Port:                 int64(port),
+			DestinationNamespace: "",
```
aduminuc, line 51: How do we capture which GW reported it? If the uniqueness is guaranteed by the namespace+Name+Port, what happens if 2 gateways are reporting the APIs with same name and port? 
I think we either "exploit" the namespace to capture it or we add another field.
To understand which gw is reporting either you add an explicit source in the call or rely on the authentication (when will be implemented), I would go for the first for now.

api/swagger.yaml
```diff
@@ -800,6 +804,35 @@ paths:
         default:
           $ref: '#/responses/UnknownError'
 
+  /control/newDiscoveredAPIs:
+    post:
```
aduminuc, line 808: How do you know who is the notifier?

klyr: The 'source' field is here for that.

aduminuc (approved):

PR #275, opened 2022-10-25
Adds Gateway Management APIs

klyr:


klyr:
I still need to convert the swagger file to OpenAPI v3

api/swagger.yaml
```diff
@@ -833,6 +858,73 @@ paths:
...
+          $ref: '#/responses/UnknownError'
+    post:
+      summary: 'Create a new gateway'
+      parameters:
+        - in: 'body'
+          name: 'body'
+          required: true
+          schema:
+            description: 'Create a new gateway'
+            $ref: '#/definitions/APIGateway'
```
aduminuc, line 888: Can the client provide a Gateway with ID and token, or those always generated by APIClarity? 
In that case, shall we create an object like APIGateway_Create? I am not sure what's the common practise in APIclarity today

klyr: The client can, it will be validated, but it will be ignored when creating the entry.
So, APIClarity generates the ID and the Token regardless.

aduminuc (approved):

PR #295, opened 2022-11-15
add new APISampling table

thelasttoto:


backend/pkg/database/tracesampling.go
```diff
@@ -0,0 +1,77 @@
...
+	traceSamplingTableName = "trace_sampling"
+)
+
+type TraceSampling struct {
+	gorm.Model
+
+	APIID         uint   `json:"api_id,omitempty" gorm:"column:api_id" faker:"-"`
+	TraceSourceID uint   `json:"trace_source_id,omitempty" gorm:"column:trace_source_id" faker:"-"`
+	Component     string `json:"component,omitempty" gorm:"column:component" faker:"-"`
+}
```
klyr, line 32: Need to add a unique index constraint on (APIID, TraceSourceID, Component)

klyr: Should we add a TraceSourceID as a foreign key ?

thelasttoto: No Foreign key here, as I can have TraceSourceID==0 as valid value

thelasttoto: Just seeing your PR 299... Now TraceSourceID==0 is a valid source. BUT:
- in other place (for ex apiinventory.go) TraceSourceID is not a foreign key,
- also... event_id, api_id have never foreign key. 
What missed, if any, is to delete entries from trace_sampling if a TraceSource is deleted

klyr (approved):
Let's go !

PR #298, opened 2022-11-21
allow APIClarity to start without kubernetes

jnapper7:
This PR provides a new config flag to start APIClarity without kubernetes.

  * A new boolean env var "ENABLE_K8S" is checked at boot.
  * The env var "TRACE_SAMPLING_ENABLED" configuration is enforced by disabling the trace-sampling-manager.

By default the env var *ENABLE_K8S* is set to __True__, allowing backwards-compatible behavior with existing deployments.

Since the trace-sampling-manager still requires k8s to store some data, the PR also includes a change to disable the trace sampling manager when *TRACE_SAMPLING_ENABLED* is __False__. This may cause backward-compatibility problems as the behavior can be different in edge cases.

So, realistically, to boot APIClarity cleanly without k8s, you must use the following config and this PR:
  * ENABLE_K8S=False
  * TRACE_SAMPLING_ENABLED=False

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,6 +182,7 @@ Update [values.yaml](https://github.com/openclarity/apiclarity/blob/master/chart
    DATABASE_DRIVER=LOCAL K8S_LOCAL=true FAKE_TRACES=true FAKE_TRACES_PATH=./backend/pkg/test/trace_files \
    ENABLE_DB_INFO_LOGS=true ./backend/bin/backend run
    ```
+   Note: this command requires a proper KUBECONFIG in your environment when __K8S_LOCAL=true__ is used. If you want to run without k8s, use __ENABLE_K8S=false__ instead.
 
 4. Open APIClarity UI in the browser: <http://localhost:8080/>
 
--- a/backend/cmd/backend/main.go
+++ b/backend/cmd/backend/main.go
@@ -56,6 +56,7 @@ func main() {
 	viper.SetDefault(config.DatabaseCleanerIntervalSec, "30")
 	viper.SetDefault(config.StateBackupFileName, "state.gob")
 	viper.SetDefault(config.DatabaseDriver, database.DBDriverTypePostgres)
+	viper.SetDefault(config.EnableK8s, true)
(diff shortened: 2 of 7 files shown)


fishkerez:
Hey @jnapper7 
Out of curiosity, what is the difference between ENABLE_K8S and K8S_LOCAL? 
Currently I can run Apiclarity locally with the command from the README:
```
DATABASE_DRIVER=LOCAL K8S_LOCAL=true FAKE_TRACES=true FAKE_TRACES_PATH=./backend/pkg/test/trace_files \
ENABLE_DB_INFO_LOGS=true ./backend/bin/backend run
```

So why do I need also ENABLE_K8S? And should it be stated in the README?

jnapper7:
> Hey @jnapper7 Out of curiosity, what is the difference between ENABLE_K8S and K8S_LOCAL? Currently I can run Apiclarity locally with the command from the README:
> 
> ```
> DATABASE_DRIVER=LOCAL K8S_LOCAL=true FAKE_TRACES=true FAKE_TRACES_PATH=./backend/pkg/test/trace_files \
> ENABLE_DB_INFO_LOGS=true ./backend/bin/backend run
> ```
> 
> So why do I need also ENABLE_K8S? And should it be stated in the README?

It's a good question. However, K8S_LOCAL does not turn off the requirement for kubernetes. If you check the code:

/backend/pkg/k8smonitor/common.go/CreateLocalK8sClientset

you can see that the reason it starts is that you have a proper KUBECONFIG in your environment. If you remove that, then APIClarity will not boot. This PR is to fix that by disabling dependencies on k8s entirely.

fishkerez:
I see, that make sense.
I think that in that case, the current instructions should be stayed as is in the README, since it brings more functionality. 
But maybe worth adding a "Note: this command require a proper KUBECONFIG in your environment. If you want to run without k8s, use ENABLE_K8S=false"
WDYT?

jnapper7:
Good idea! I added your text with a little bit extra explanation. :)

aduminuc (approved):
Tested this version in panoptica and work. 
LGTM

PR #314, opened 2022-12-20
Trace sources support

klyr:
Adds Trace Sources support.
* A Trace Source is the entity which is sending traces to APIClarity (A wasm filter, or an external (to Kubernetes) APIGateway, like Kong or Apgee-X.
* A Service is now also identified by its trace source.
* Adds support for creation/deletion of trace sources, via API, no UI yet.
* Adds more foreign key constraints to the database, and cascade delete when trace source or services are deleted.
* When creating a new trace source, generate an api token for this trace source. This token is then sent along the trace in order to identify the trace source which created the trace.
* The API token is only usuable when sending traces to the new https only server.


thelasttoto (approved):
Go!

PR #320, opened 2023-01-06
changes for standalone loadbalancer support for helm chart

pkalum:
This is the changes for the standalone loadbalancer service type of external APIClarity service

### DIFF
--- a/charts/apiclarity/values.yaml
+++ b/charts/apiclarity/values.yaml
@@ -46,7 +46,7 @@ apiclarity:
   # notificationURLPrefix: example.local:8888/
 
   tls:
-    enabled: false
+    enabled: true
     # Secret that contains server tls key and cert
     tlsServerCertsSecretName: "apiclarity-tls"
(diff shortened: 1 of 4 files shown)


charts/apiclarity/values.yaml
```diff
@@ -46,7 +46,7 @@ apiclarity:
   # notificationURLPrefix: example.local:8888/
 
   tls:
-    enabled: false
+    enabled: true
```
klyr, line 49: You should leave the default to false

charts/apiclarity/values.yaml
```diff
@@ -267,3 +267,11 @@ APIClarityRuntimeFuzzing:
 
 # End of APIClarity Runtime Fuzzing Values
 #######################################################################################
+
+#######################################################################################
+## APIClarity External Trace Source Values
+
+supportExternalTraceSource:
+  enabled: true
```
klyr, line 275: In my opinion, it should default to false

PR #326, opened 2023-01-16
Update kong plugin

thelasttoto:
Create new ApiClarity client to be used by go plugin
ApiClarity client relying on new apiclarity gateway interface 
Update kong plugin to use new ApiClarity client


plugins/common/utils.go
```diff
@@ -151,7 +151,8 @@ func createClientTransportTLS(host string, tlsOptions *ClientTLSOptions) (runtim
 
 	//Trust the augmented cert pool in our client
 	tlsConfig := &tls.Config{
-		RootCAs: rootCAs,
+		RootCAs:            rootCAs,
+		InsecureSkipVerify: false,
```
klyr, line 155: Isn't it already the default value ?

plugins/gateway/kong/plugin.go
```diff
@@ -17,48 +17,66 @@ package main
...
-		// TODO tls will not work since trace sampling manager is not supporting it currently
-		traceSampling, err := trace_sampling_client.Create(false, conf.TraceSamplingHost, common.SamplingInterval)
+	if conf.TraceSamplingEnabled && apiclarityClient == nil {
+		_ = kong.Log.Info(fmt.Sprintf("Host: '%v'", conf.Host))
+		_ = kong.Log.Info(fmt.Sprintf("TraceSamplingEnabled: '%v'", conf.TraceSamplingEnabled))
+		_ = kong.Log.Info(fmt.Sprintf("Token: '%v'", conf.Token))
+		_ = kong.Log.Info(fmt.Sprintf("SamplingInterval: '%v'", common.SamplingInterval))
+		discoveredApis = []string{}
+		if conf.EnableTLS {
+			if _, err := os.Stat("/etc/traces/certs"); os.IsNotExist(err) {
```
klyr, line 64: "/etc/traces/certs" should go in the common utils.

plugins/gateway/kong/plugin.go
```diff
@@ -78,50 +105,69 @@ func (conf Config) Response(kong *pdk.PDK) {
...
+		return fmt.Errorf("failed to get routed service for processNewDiscoveredAPI: %v", err)
+	}
+	// Check for newDiscoveredApi
+	if !common.Contains(discoveredApis, host) {
+		appendNewDiscoveredAPI(host)
+		hosts := []string{host}
+		if err := apiclarityClient.PostNewDiscoveredAPIs(hosts); err != nil {
+			return fmt.Errorf("failed to send newDiscoveredApi request: %v", err)
+		}
+		_ = kong.Log.Info("Sent PostNewDiscoveredAPIs with success")
```
klyr, line 134: Should be logged only in Debug mode

plugins/gateway/kong/plugin.go
```diff
@@ -78,50 +105,69 @@ func (conf Config) Response(kong *pdk.PDK) {
...
-		return false, fmt.Errorf("failed to get routed service: %v", err)
+		return false, fmt.Errorf("failed to get routed service for shouldTrace: %v", err)
 	}
 	host, port, _ := parseKongHost(routedService.Host)
-	if traceSamplingClient.ShouldTrace(host, port) {
+	// Note: here 'host' contain already the namespace. i.e. for service 'catalog' on namespace 'sock-shop', host will be 'catalog.sock-shop'
+	if apiclarityClient.ShouldTrace(host, port) {
 		return true, nil
 	}
 	_ = kong.Log.Info("Ignoring host: %v:%v", host, port)
```
klyr, line 156: Debug mode, I know, it was already there.

plugins/gateway/kong/plugin.go
```diff
@@ -190,11 +236,14 @@ func createTelemetry(kong *pdk.PDK) (*models.Telemetry, error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed to get response headers: %v", err)
 	}
-	host, port, namespace := parseKongHost(routedService.Host)
+	host, port, _ := parseKongHost(routedService.Host)
 
+	// Note by Axel. Currently, we populate Host with (host+namespace), and don't put namespace on DestinationNamespace because
+	// on an issue on the way we create a service in ApiClarity either with newDiscoveredApi either with Telemetry
+	// Must be fixed with PA-11655
```
klyr, line 243: Remove that comment

plugins/gateway/kong/plugin.go
```diff
@@ -17,48 +17,66 @@ package main
...
-	if conf.TraceSamplingEnabled && traceSamplingClient == nil {
-		_ = kong.Log.Info("Creating trace sampling client")
-		// TODO tls will not work since trace sampling manager is not supporting it currently
-		traceSampling, err := trace_sampling_client.Create(false, conf.TraceSamplingHost, common.SamplingInterval)
+	if conf.TraceSamplingEnabled && apiclarityClient == nil {
+		_ = kong.Log.Info(fmt.Sprintf("Host: '%v'", conf.Host))
+		_ = kong.Log.Info(fmt.Sprintf("TraceSamplingEnabled: '%v'", conf.TraceSamplingEnabled))
+		_ = kong.Log.Info(fmt.Sprintf("Token: '%v'", conf.Token))
+		_ = kong.Log.Info(fmt.Sprintf("SamplingInterval: '%v'", common.SamplingInterval))
+		discoveredApis = []string{}
```
klyr, line 62: Maybe having a map could be faster, no need to iterate over the list for each trace

backend/pkg/traces/server.go
```diff
@@ -210,6 +210,7 @@ func (s *HTTPTracesServer) getHostsToTrace(params operations.GetHostsToTracePara
 }
 
 func (s *HTTPTracesServer) newDiscoveredAPIs(params operations.PostControlNewDiscoveredAPIsParams) middleware.Responder {
+	log.Infof("newDiscoveredAPIs was invoked")
```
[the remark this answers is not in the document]
thelasttoto: I added this trace because the corresponding one exists in control_controller.go for PostControlNewDiscoveredAPIs. During the tests I expected a such trace in log without seeing it until I realized the diff. Then I added it for newDiscoveredAPIs to have the same behavior.
But I can remove it.

plugins/common/apiclarity_client/apiclarity_client.go
```diff
@@ -0,0 +1,140 @@
+// Copyright © 2021 Cisco Systems, Inc. and its affiliates.
```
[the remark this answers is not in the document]
thelasttoto: ok, I update this. But be warned, 2021 is used everywhere else.

plugins/common/utils.go
thelasttoto: yes. I can remove it.

plugins/gateway/kong/plugin.go
```diff
@@ -17,48 +17,66 @@ package main
...
 
 func (conf Config) Access(kong *pdk.PDK) {
-	if conf.TraceSamplingEnabled && traceSamplingClient == nil {
-		_ = kong.Log.Info("Creating trace sampling client")
-		// TODO tls will not work since trace sampling manager is not supporting it currently
-		traceSampling, err := trace_sampling_client.Create(false, conf.TraceSamplingHost, common.SamplingInterval)
+	if conf.TraceSamplingEnabled && apiclarityClient == nil {
+		_ = kong.Log.Info(fmt.Sprintf("Host: '%v'", conf.Host))
+		_ = kong.Log.Info(fmt.Sprintf("TraceSamplingEnabled: '%v'", conf.TraceSamplingEnabled))
+		_ = kong.Log.Info(fmt.Sprintf("Token: '%v'", conf.Token))
```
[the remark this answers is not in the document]
thelasttoto: ok

plugins/gateway/kong/plugin.go
thelasttoto: hmmm... why? it's part of the template to patch kong plugin and we can't assume that all plugin will use the same mount path...
But for sure I can put this in a constant or in configuration with default value. 
The goal here is simply to verify that the certificate mount path exists.

plugins/gateway/kong/plugin.go
thelasttoto: Ok. But anyway, info or debug are not display on logs. only error appear.

plugins/gateway/kong/plugin.go
thelasttoto: You talk about the _ = kong.Log.Info("Ignoring host: %v:%v", host, port), right? Ok

plugins/gateway/kong/plugin.go
thelasttoto: hmmm... then in a general way, we need to define what kind of comments we want to put on the code: Here it is a comment that can help a future developer who will have to update this part of the code and can ask itself why namespace is not used here.
But I can remove if needed. (for me it's a kind of obfuscation)

plugins/gateway/kong/plugin.go
thelasttoto: I put these during implementation and keep it because it can help to see in log what is the config used to start the plugin. I can remove all of this if it disturb (it was not present before)

plugins/api/swagger.yaml
```diff
@@ -144,6 +157,27 @@ definitions:
...
+      message:
+        type: 'string'
+
+  InternalErrorResponse:
+    description: 'An object that is return in all cases of failures.'
+    type: 'object'
+    properties:
+      message:
+        type: 'string'
+
```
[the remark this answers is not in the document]
thelasttoto: I will remove this.
But for information, with openapispec_gen source code I encountered errors related to the use of
        default:
in response of the telemetry request.
It's why I defined properly some common errors.

plugins/gateway/kong/plugin.go
thelasttoto: ok, I see what you mean.  constant CACertFile = "/etc/traces/certs/root-ca.crt" exists in common

plugins/gateway/kong/plugin.go
klyr: No, I'm talking about `discoveredApis = []string{}` It's currently a list, and if I understand correctly, it's used to keep tracks of discovered apis, so, basically, it's just a set.
It was a suggestion to use a set instead of a list.
But at the end, the size of this list will never be very big, so iterrating on each element on each received traces won't take long.
As you prefer.

plugins/gateway/kong/plugin.go
klyr: I'm just trying to avoid to many log output when everything is working well.

aduminuc (commented):
This looks good to me, however before merging I would like to make sure that the deploy would work for panoptica as well.

To do that we need to finish the other tasks to properly integrate kong in oreo. So I would propose to keep this in a branch until that stuff is ready.

